PR #20, opened 2019-08-07
Hash context

SebCorbin:
* [x] HOC (`withHashParameters`)
  * [x] `parameters` (paramètres de hash à surveiller)
  * [x] `getHashParameters()` prend la config depuis le hash et la renvoie (filtrée sur `parameters`)
  * [x] `setHashParameters()` va chercher la valeur de `parameters` depuis le state/valeurs et l'injecte dans le hash
* [x] L'état du tableau
* [x] Pas l'affichage du panneau de filtre
* [ ] ~~Les filtres~~ https://github.com/Terralego/terra-front/issues/21
* [ ] ~~L'affichage ou non de la mini-fiche~~ https://github.com/Terralego/terra-front/issues/22
* [ ] ~~L'objet sélectionné~~ https://github.com/Terralego/terra-front/issues/22
* [x] L'état d'affichage (visible, plié) de l'arbre des couches

cf https://gitlab.makina-corpus.net/Terralego/Core/project/wikis/meetings/20190718_permalinks


src/modules/Visualizer/services/layersTreeUtils.js
```diff
@@ -4,18 +4,36 @@ export const INITIAL_FILTERS = new Map();
...
-export function initLayersStateAction (layersTree) {
+/**
+ * Returns a flattened map of layers state from a layers tree
+ *
+ * @param {object} layersTree The layers tree config object
+ * @param {string|string[]} layers Active layers from hash (comma-separated)
+ * @param {string} table Active table from hash (layer id)
+ * @return {Map} A reduced layer tree state
+ */
+export function initLayersStateAction (layersTree, { layers, table } = {}) {
```
mabhub: Could you make it as an Arrow function instead?

```suggestion
export const initLayersStateAction = (layersTree, { layers, table } = {}) => {
```

src/modules/Hash/withHashParameters.js
```diff
@@ -0,0 +1,87 @@
...
+     */
+    setHashParameters = values => {
+      // Filter hash values on parameters
+      const params = {
+        ...parse(window.location.hash, this.options),
+        ...this.getFilteredParams(values),
+      };
+
+      // Rebuild string and set hash
+      window.history.replaceState(window.history.state, '', `#${stringify(params, this.options)}`);
```
jrmi: I think we shouldn't force the URL update here. We may want to generate a link on demand. If we keep the automatic updated, it should be in a higher level, don't you think ?

src/modules/Visualizer/LayersTree/LayersTreeProvider/LayersTreeProvider.js
```diff
@@ -122,17 +128,32 @@ export class LayersTreeProvider extends React.Component {
...
   resetState (state, callback = () => {}) {
+    const { setHashParameters } = this.props;
+
     this.setState(state, () => {
       callback();
       const { onChange } = this.props;
       const { layersTreeState } = this.state;
 
+      // Simplify the state from the map, and send it to hash
+      const activeLayers = [];
```
jrmi: I think we should have no code concerning hash management in any other component as we discuss. The component should send the object as is and let the HOC hash manager do his job. Otherwise we are going to have repeated boilerplate code in multiples components and tightly coupled components.

src/modules/Hash/withHashParameters.js
```diff
@@ -0,0 +1,87 @@
...
+    };
+
+    render () {
+      const { ...props } = this.props;
+      const { setHashParameters, getHashParameters } = this;
```
mabhub: There is no real benefit of destructuring `this`.

It is preferable to call these methods through `this` object instead of destructuring.

Generally, it avoid losing instance binding inside the method, but it also help identifying (and readability) _methods_ as opposite to  _functions_.

src/modules/Hash/withHashParameters.js
SebCorbin: Only the state is changed, `hashupdate` browser event is not really called (I took the code from mapbox, with also listens to `hashupdate`. This event is only called when the user manually change the url.

My initial thought is: copy-pasting fro adressbar is a native user reflex, and we should not force them to hit a button to get the permalink as long as we can.

jrmi: Too long URL can be difficult to copy paste. Also users may don't want to keep the current context but keep the initial url/context. Sharing a fixed view should be a active action. That's also what we decide during the meeting.

Considering technical aspect, i think the hash update should be a separate component. Remember this component can be reused in another context/application and have another usage. Here you constraint the usage.

src/modules/Visualizer/LayersTree/LayersTreeProvider/LayersTreeProvider.js
SebCorbin: There is nothing concerning hash management per se, it's only working a hash state, which we need because there is already an `initialState` coming from the config and a `layerTreeState`. 

Also, getting any kind of object from hash without sanitation exposes us to all kind of future bugs we don't want to process.

SebCorbin: Also, see your your other comment:
> Too long URL can be difficult to copy paste.

Yes, so a component need to put as little as possible in the hash, thus is needs to pre-process it.

src/modules/Hash/withHashParameters.js
SebCorbin: Agreed, I'll put another component just for setting hash.

Just for the record, some components like mapbox are not as thoughtful and just update the hash without asking.

jrmi: That's why you need to create a merge request to have the possibility to use the hash another way ;-)

src/modules/Visualizer/LayersTree/LayersTreeProvider/LayersTreeProvider.js
jrmi: > There is nothing concerning hash management per se, it's only processing a hash state, 

I rephrase, we should have no code concerning hash state here ;-)

> Also, getting any kind of object from hash without sanitation exposes us to all kind of future bugs we don't want to process.

Which one ?

src/modules/Hash/withHashParameters.js
jrmi: Quote from misplaced [comments](https://github.com/Terralego/terra-front/pull/20#discussion_r311960051)

> Yes, so a component needs to put as little as possible in the hash, thus is needs to pre-process it.

Agree with that but we are going to have long url very fast.

src/modules/State/Hash/withHashState.js
```diff
@@ -0,0 +1,52 @@
...
+ * @param {bool} listenHash Reload initialSate when hash is changed
+ */
+export const withHashState = ({ updateHash, listenHash } = {}) => WrappedComponent =>
+  class WithHashState extends React.Component {
+    options = DEFAULT_OPTIONS;
+
+    constructor (props) {
+      super(props);
+      if (listenHash) {
+        window.addEventListener('hashchange', this.forceUpdate, false);
```
mabhub: As you add an _eventListenner_, you have to remove it sooner or later: if component is unmounted, event may be triggered, then React will throw a Warning as you ask to `forceUpdate` an unmounted component.

image

Usually, event binding is done in `componentDidMount()`, and unbinding in `componentWillUnmount()`.

src/modules/State/Hash/withHashState.js
```diff
@@ -0,0 +1,73 @@
+import * as PropTypes from 'prop-types';
```
mabhub:
```suggestion
import PropTypes from 'prop-types';
```

SebCorbin: Screenshot 2019-08-12 at 16 33 56

src/modules/Hash/withHashParameters.js
SebCorbin: Refactored

src/modules/State/Hash/withHashState.js
SebCorbin: Reverted, in fact it's an issue with Jetbrains https://intellij-support.jetbrains.com/hc/en-us/community/posts/360003783160-React-Native-prop-types-module-Default-export-is-not-declared-in-imported-module- and `react-contenteditable` https://github.com/lovasoa/react-contenteditable/issues/171

PR #36, opened 2019-08-19
Fix optim interactions

hadrienl:
https://github.com/Terralego/terra-front/issues/37

Interactions take lot of cpu because of layers check. We get perfs by memoizing the list of layers and refreshing it when all layers list change.

### DIFF
--- a/src/modules/Map/services/mapUtils.js
+++ b/src/modules/Map/services/mapUtils.js
@@ -1,15 +1,14 @@
 import bbox from '@turf/bbox';
 
-import moize from 'moize';
 import { PREFIXES } from './cluster';
 
 const PREV_STATE = {};
 
-export const getLayers = moize((map, layerId) => {
+export const getLayers = (map, layerId) => {
   const regexp = new RegExp(`^${layerId}(-(${PREFIXES.join('|')}))?(-[0-9]+)?$`);
   return map.getStyle().layers
     .filter(({ id }) => id.match(regexp));
-});
+};
 
 export function toggleLayerVisibility (map, layerId, visibility) {
   getLayers(map, layerId)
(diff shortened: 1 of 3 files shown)


src/modules/Map/services/mapUtils.js
```diff
@@ -83,94 +85,120 @@ export const checkContraints = ({
...
     map.on(eventType, id, event => {
-      callback({
-        event,
-        map,
-        layerId: id,
-        interaction,
-        feature: PREV_STATE.point,
-        eventType,
-      });
+      const features = map.queryRenderedFeatures(PREV_STATE.point);
```
SebCorbin: When specifying a layer id, you get the features in the event, there's no need to query twice

src/modules/Map/services/mapUtils.js
```diff
@@ -83,94 +85,120 @@ export const checkContraints = ({
...
+
+  let interactions = false;
+
+  features.some(feature => {
+    const { layer: { id: layerId } } = feature;
+
+    const foundInteractions = eventInteractions.filter(({ id, trigger = 'click', constraints }) => {
+      const found = getLayers(map, id).find(({ id: compatibleId }) =>
+        layerId === compatibleId &&
+        map.getLayoutProperty(compatibleId, 'visibility') === 'visible');
```
SebCorbin: There's no need to do that, as `queryRenderedFeatures` only return features of visible layers

src/modules/Map/services/mapUtils.js
```diff
@@ -83,94 +85,120 @@ export const checkContraints = ({
...
+  eventType,
+  map,
+  point,
+  interactions: eventInteractions,
+}) {
+  let features;
+  try {
+    features = map.queryRenderedFeatures(point);
+  } catch (e) {
+    return null;
```
SebCorbin: A comment on why this would fail would be nice

src/modules/Map/services/mapUtils.js
```diff
@@ -83,94 +85,120 @@ export const checkContraints = ({
...
+      const found = getLayers(map, id).find(({ id: compatibleId }) =>
+        layerId === compatibleId &&
+        map.getLayoutProperty(compatibleId, 'visibility') === 'visible');
+
+      if (!found || (constraints && !checkContraints({ map, constraints, feature }))) {
+        return false;
+      }
+
+      return found &&
+        eventType === (trigger === 'mouseover' ? 'mousemove' : trigger);
```
SebCorbin: I would put this simple check long before checking that a features belong to a layer (or its siblings) for perf improvements.

src/modules/Map/services/mapUtils.js
hadrienl: I don't know why but sometimes it does.

mabhub: Silent error is not the solution. At least a `console.error()` or `console.info()` should expose issue for dev purpose.

src/modules/Map/services/mapUtils.js
hadrienl: I do not have features in the event. But Looking for it helps me to find that clusters layers where not all covered by mouseleave event.

src/modules/Map/services/mapUtils.js
```diff
@@ -1,11 +1,13 @@
 import bbox from '@turf/bbox';
-
 import moize from 'moize';
+
 import { PREFIXES } from './cluster';
 
 const PREV_STATE = {};
 
-export const getLayers = moize((map, layerId) => {
+export const getLayers = moize({
```
SebCorbin: Missing documentation

src/modules/Map/services/mapUtils.js
```diff
@@ -83,94 +85,120 @@ export const checkContraints = ({
...
+  });
+
+  return interactions;
+}
 
+const listenerWaiter = {};
 export function setInteractions ({ map, interactions, callback }) {
   const eventsTypes = new Set(interactions.reduce((triggers, { trigger = 'click' }) => [...triggers, trigger], []));
 
-  // 'mouseover' is the trigger, 'mousemove' is the eventType
```
SebCorbin: I would document (here or elsewhere), how mousemove is the same as 'mouseover', or why we replace it

src/modules/Map/services/mapUtils.js
SebCorbin: Wouldn't it be faster to store the features from the last PREV_STATE.point instead of querying it again?

src/modules/Map/services/mapUtils.js
hadrienl: This is not the scope of this PR.

src/modules/Map/services/mapUtils.js
hadrienl: No, because when we store `PREV_STATE.point` we don't query features and may not query it later.

SebCorbin: You query features just after setting it, in `getInteractionsOnEvent`

src/modules/Map/services/mapUtils.js
```diff
@@ -83,94 +85,119 @@ export const checkContraints = ({
...
 
-let listenerWaiter = 0;
+export function getInteractionsOnEvent ({
+  eventType,
+  map,
+  point,
+  interactions: eventInteractions,
+  features = map.queryRenderedFeatures(point),
+}) {
+  let interactions = false;
```
jrmi: This default value let the reader thinks that `interactions` is a boolean value. Can it be a `null` (or undefined ) instead?
```suggestion
  let interactions = null;
```

src/modules/Map/services/mapUtils.js
```diff
@@ -83,94 +85,119 @@ export const checkContraints = ({
...
+  features.some(feature => {
+    const { layer: { id: layerId } } = feature;
+
+    const foundInteractions = eventInteractions.filter(({ id, trigger = 'click', constraints }) => {
+      if (eventType !== (trigger === 'mouseover' ? 'mousemove' : trigger)) return false;
+
+      const found = getRelatedLayers(map, id)
+        .find(({ id: compatibleId }) => layerId === compatibleId);
+
+      return found && (!constraints || checkContraints({ map, constraints, feature }));
```
mabhub:
```suggestion
      if (!found) {
        return false;
      }
      
      if (!constraints) {
        return true;
      }
      
      return checkContraints({ map, constraints, feature });
```

src/modules/Map/services/mapUtils.js
hadrienl: I don't really want to refacto everything in this file. That's not a part of code which take parts of this PR but some code which have already been reviewed and merged weeks ago.

PR #42, opened 2019-08-22
Add layers ordered by type

hadrienl:


src/modules/Map/Map/Map.js
```diff
@@ -273,7 +275,12 @@ export class MapComponent extends React.Component {
     const { map, customStyle: { sources = [], layers = [] } } = this.props;
     const { layers: allLayers } = map.getStyle();
     sources.forEach(({ id, ...sourceAttrs }) => map.addSource(id, sourceAttrs));
-    layers.forEach(layer => {
+    const orderedLayers = [...layers];
+    orderedLayers.sort(({ type: typeA }, { type: typeB }) => {
+      if (typeA === typeB) return 0;
```
mabhub: I'm not sure comparing `typeA` & `typeB` before looking for their indexes in `LAYERS_TYPES` improves perfs significativelly.

As indexes difference would be `0` when types are the same, you could simplify this way:

```js
    orderedLayers.sort(({ type: typeA }, { type: typeB }) =>
        (LAYERS_TYPES.indexOf(typeA) - LAYERS_TYPES.indexOf(typeB)));
```

hadrienl: `if (typeA === typeB) return 0;` saves two indexOf. It's not a lot but it's better perf without losing lisibility.

PR #56, opened 2019-08-29
Fix translation mapbox control

DarkVisenya:
Add translation in navigation control and fix translation in permalink control

### DIFF
--- a/src/modules/Map/Map/Map.js
+++ b/src/modules/Map/Map/Map.js
@@ -425,19 +425,15 @@ export class MapComponent extends React.Component {
           map.addControl(controlInstance, position);
 
           const { _container: container } = controlInstance;
-          const isNavigationControlGroup = container.querySelector('.mapboxgl-ctrl-zoom-in');
 
-          if (isNavigationControlGroup) {
+          if (control === CONTROL_NAVIGATION) {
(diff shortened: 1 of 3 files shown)


src/modules/Map/Map/Map.js
```diff
@@ -405,6 +411,16 @@ export class MapComponent extends React.Component {
           const { disabled } = params;
           this.controls.push(controlInstance);
           map.addControl(controlInstance, position);
+
+          const { _container: container } = controlInstance;
+          const isNavigationControlGroup = container.querySelector('.mapboxgl-ctrl-zoom-in');
```
mabhub:
```suggestion
          const isNavigationControlGroup = container.querySelector('.mapboxgl-ctrl-group');
```

DarkVisenya: I can't do that. 
The other element like PermalinkControl or CaptureControl have alse this class

src/modules/Map/Map/Map.test.js
```diff
@@ -527,6 +531,7 @@ describe('controls', () => {
       // eslint-disable-next-line class-methods-use-this
       onRemove () {}
     }
+    // instance.props.translate = jest.fn(() => 'foo');
```
mabhub: ?

PR #57, opened 2019-08-30
Regroup Storybook components

mabhub:


src/stories/modules/Map/index.js
```diff
@@ -3,7 +3,7 @@ import { storiesOf } from '@storybook/react';
 import MapWithCluster from './MapWithCluster';
 import SyncedMaps from './SyncedMaps';
 
-const stories = storiesOf('Modules/Map/', module);
+// const stories = storiesOf('Modules/Map/', module);
```
hadrienl: delete

src/stories/modules/Table/Table.stories.js
```diff
@@ -23,7 +23,7 @@ const data = [
   ['lorem', 'ipsum', '12345.67890'],
 ];
 
-storiesOf('Modules/Table/', module)
+storiesOf('Components/Table', module)
```
hadrienl: You should keep "Module" to help know in which module the component is from.

mabhub: That was the point of decision taken with @jrmi and the main subject of this PR: take each component out of modules.

PR #65, opened 2019-09-02
Fix print control

hadrienl:


src/modules/Map/Map/components/PrintControl/styles.scss
```diff
@@ -1,7 +1,6 @@
 .visualizer {
   .visualizer__print {
     margin: 0 auto;
-    position: relative;
```
hadrienl: This line made print view break after the tooltip zindex fix.

PR #66, opened 2019-09-02
Translate placeholder of select filter

SebCorbin:


src/modules/Forms/Filters/Filters.js
```diff
@@ -86,6 +87,7 @@ export class Filters extends React.Component {
     layer: '',
     translate: translateMock({
       'terralego.form.filter.select_placeholder': 'Enter a query first',
+      'terralego.form.filter.input_placeholder': 'Filter...',
```
hadrienl: Module name is 'forms' with a 's'. Could you use the same name for the keys ?

src/modules/Forms/Filters/index.js
```diff
@@ -1,4 +1,5 @@
+import { connectLayersTree } from '../../Visualizer/LayersTree/LayersTreeProvider/context';
 import Filters from './Filters';
 
 export { TYPE_SINGLE, TYPE_MANY, TYPE_RANGE, TYPE_BOOL } from './Filters';
-export default Filters;
+export default connectLayersTree('translate')(Filters);
```
hadrienl: You cannot link components with provider from other modules. The translate prop here can only be passed by props.

src/modules/Auth/components/LoginForm/LoginFormRenderer.js
```diff
@@ -30,6 +30,7 @@ export const LoginFormRenderer = ({
       onSubmit={submit}
     >
       <FormGroup
+        // i18next-extract-disable-next-line
```
hadrienl: To avoid this kind of comment, you could use `context` instead of dynamic key:

```
translate('auth.loginform.email', { context: errorLogin ? 'invalid' : 'help'})
```

will get "auth.loginform.email_invalid" or "auth.loginform.email_help" key (and fallback to "auth.loginform.email" if context value doesn't exist.

hadrienl: btw… I don't see any trace of i18next-extract in codebase. Why this comment ? This library should be agnostic for translation lib and shouldn't be hard linked to i18next.

locales/fr/translation.json
```diff
@@ -38,8 +38,27 @@
   },
```
hadrienl: Oh I didn't see this file were added. It shouldn't. It's a i18next format file and translate should be agnostic. You should replace this file by a .md file with list of all available translations.

src/modules/Auth/components/LoginForm/LoginFormRenderer.js
SebCorbin: This is a comment used by the extractor, the library is still agnostic, but the extractor can be used in development.

locales/fr/translation.json
SebCorbin: Translate is still agnostic, there is no import of i18next anywhere, it’s just a helper for projects to import common translations, see https://github.com/Terralego/terra-front/pull/74

src/modules/Auth/components/LoginForm/LoginFormRenderer.js
hadrienl: You shouldn't keep it.

locales/fr/translation.json
hadrienl: Ok

PR #73, opened 2019-09-03
Remove autofocus on twitter share control button

MarianneLa:


src/modules/Map/Map/components/ShareControl/ShareControl.js
```diff
@@ -152,10 +152,9 @@ export class ShareControl extends AbstractMapControl {
               content={translate('terralego.map.share_control.share', {
                 context: network.charAt(0).toUpperCase() + network.slice(1),
               })}
+              openOnTargetFocus={false}
             >
-              <Button
-                onClick={this.share(network)}
```
hadrienl: Why did you remove the onClick ?

MarianneLa: mistake...

PR #76, opened 2019-09-04
Fix iterator (duplicate) key issue in LayersTree

mabhub:


src/modules/Visualizer/LayersTree/LayersTree.js
```diff
@@ -9,21 +9,24 @@ import './styles.scss';
...
-      ? (
-        <LayersTreeGroup
-          key={layer.group}
-          title={layer.group}
-          layer={layer}
-        />
-      )
-      : (
+    {layersTree.map(layer => {
+      if (layer.group && !layer.exclusive) {
```
hadrienl: Why didn't you keep the ternary form ?

mabhub: Ternary operator wrecks readability for multiline JSX and should be avoid.

Ternary operator is good for single line condition, or when each argument could be a single line
```js
condition ? exprIfTrue : exprIfFalse;

// or

condition
  ? exprIfTrue
  : exprIfFalse;

// but not

condition
  ? (
    // multiline exprIfTrue
  )
  : exprIfFalse;
```

Issue #80, opened 2019-09-04
Range filter doesn't update its value


SebCorbin:
The value is updated onRelease, but its value should also be update onChange, without notifying search

SebCorbin:
This is in part due to the complex `ContentEditable` handling, this would likely be fixed by #84

PR #87, opened 2019-09-05
Revamp Range filter

### LINKED ISSUE #84: UX: Range filter is unusable with lot of steps
Proposal: when number of steps > 250 (or a configurable value), replace RangeSlider with 2 numeric input both the range.

Remove all logic about ContentEditable which is not UX-friendly and not usable in Safari

SebCorbin:
Fixes #84 #80


src/modules/Forms/Controls/Range.js
```diff
@@ -23,96 +34,123 @@ export class Range extends React.Component {
...
 
-  state = {}
-
-  onBlur = () => {
-    this.handleManualChange();
-    this.setState({ newMin: undefined, newMax: undefined });
+  constructor (props) {
+    super(props);
+    const { min = DEFAULT_MIN, max = DEFAULT_MAX, value = [min, max] } = props;
+    this.state = { range: value.length === 2 ? value : [min, max] };
```
mabhub: I'm not sure of the point of initiating state in constructor as you could just do a class property:
```js
export class Range extends React.Component {
  // static propTypes...

  // static defaultProps...

  state= {
    range: (value.length === 2 ? this.props.value : [this.props.min, this.props.max]),
  };

  // ...
```

This could avoid having this constructor method that did not anything else.

src/modules/Forms/Controls/Range.js
```diff
@@ -23,96 +34,123 @@ export class Range extends React.Component {
...
       label,
-      value,
       onChange,
       min,
       max,
-      ...props
+      inputProps,
+      translate,
+      threshold,
+      stepSize = 1,
```
mabhub: Couldn't `stepSize` default value rather be in `defaultProps` ?

src/modules/Forms/Controls/Range.js
```diff
@@ -23,96 +34,123 @@ export class Range extends React.Component {
...
-        <RangeSlider
-          value={value || [min, max]}
-          onRelease={onChange}
-          min={min}
-          max={max}
-          labelRenderer={this.labelRenderer}
-          {...props}
-        />
+        {
+          tooManyValues ? (
```
mabhub: Avoid _too-many-lines-ternary-operator_ inside JSX. Prefer this:
```jsx
<p>
  {tooManyValues && (
    // ...
  )}

  {!tooManyValues && (
    // ...
  )}
</p>
```

*(Or maybe `{Boolean(tooManyValues) && (/* ... */)}`)*

src/modules/Forms/Controls/Range.js
```diff
@@ -1,20 +1,31 @@
...
+import {
+  RangeSlider,
+  NumericInput,
+  Intent,
+  FormGroup,
+} from '@blueprintjs/core';
 import PropTypes from 'prop-types';
-import { RangeSlider } from '@blueprintjs/core';
-import ContentEditable from 'react-contenteditable';
+import React from 'react';
```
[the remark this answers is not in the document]
SebCorbin: Sorry, my editor does that sometimes...

src/modules/Forms/Controls/Range.js
```diff
@@ -23,96 +34,123 @@ export class Range extends React.Component {
...
   };
 
-  state = {}
-
-  onBlur = () => {
-    this.handleManualChange();
-    this.setState({ newMin: undefined, newMax: undefined });
+  constructor (props) {
+    super(props);
+    const { min = DEFAULT_MIN, max = DEFAULT_MAX, value = [min, max] } = props;
```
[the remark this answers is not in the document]
SebCorbin: Switched the corresponding tests to `shallow()` to avoid that

src/stories/modules/Forms/Filters.js
```diff
@@ -15,76 +21,76 @@ export class CustomFilters extends React.Component {
...
-            'UX designer': 'UX designer',
-            'Chef de projet': 'Chef de projet',
-          }, ''),
-          many_values: select('Many values with suggestions initial value', {
-            développeur: 'développeur',
-            'UX designer': 'UX designer',
-            'Chef de projet': 'Chef de projet',
-          }, ''),
-          Range_values: [number('Range from', 0), number('Range to', 100)],
+          single_value_forced: select('single value with suggestions initial value', dummyValues, ''),
```
mabhub: I know that was there before you revamped this, but I find quite weird having Storybook knobs directly in a component that is not properly inside a Storybook Story. It could result in calling knobs function outside of Storybook...

As you feel, you could push those knobs in `.stories.js` file.

src/modules/Forms/Controls/Range.js
SebCorbin: Hum, `this` would be undefined here, isn't it?

src/stories/modules/Forms/Filters.js
SebCorbin: Revamped structure
Screenshot 2019-09-05 at 18 43 46

src/modules/Forms/Controls/Range.js
hadrienl: In this specific use case, we usually add a comment to disable the linting with a comment explaining why. It's legit to disable this rule in this case.

hadrienl:
Nice refacto. Just a bemol about hooks. We decided to not use hooks in this lib because it's difficult to use it with npm link. Could you try it before merging your MR (see https://github.com/facebook/react/issues/13991#issuecomment-527289193)

SebCorbin:
> Nice refacto. Just a bemol about hooks. We decided to not use hooks in this lib because it's difficult to use it with npm link. Could you try it before merging your MR (see [facebook/react#13991 (comment)](https://github.com/facebook/react/issues/13991#issuecomment-527289193))

That's a bummer, it effectively works when I `rm -rf node_modules/{react,react-dom}` in terra-front. See https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react

What should I do, convert back to class component?

mabhub:
> That's a bummer, it effectively works when I `rm -rf node_modules/{react,react-dom}` in terra-front. See https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react
> 
> What should I do, convert back to class component?

What about just making a section in README for developpers about `npm link` and having to `rm` some dependencies?

PR #90, opened 2019-09-06
Rename the legend radius to diameter

DarkVisenya:
Be careful, this PR contains a breacking change.

Proposition to rename radius to diameter because when the svg is created, it uses the width and height attributes that correspond to the diameter

### DIFF
--- a/src/modules/Map/InteractiveMap/components/Legend/Legend.js
+++ b/src/modules/Map/InteractiveMap/components/Legend/Legend.js
@@ -8,7 +8,7 @@ import CustomComponents from './CustomComponents';
 import Circle from './components/Circle';
 import Rect from './components/Rect';
 
-const DEFAULT_RADIUS = 16;
+const DEFAULT_DIAMETER = 16;
 
 export const Legend = ({
   title,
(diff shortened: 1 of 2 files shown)


src/modules/Map/InteractiveMap/components/Legend/Legend.js
```diff
@@ -51,7 +51,7 @@ export const Legend = ({
           color,
           items: subItems,
           shape = 'square',
-          radius = DEFAULT_RADIUS,
+          diameter = DEFAULT_DIAMETER,
```
mabhub: I think you should keep backward compatibility by still allowing to have a *radius* parameter. This way, your PR won't be a breaking change anymore.

So you keep your changes about renaming and using *diameter*, but you can adjust so the app won't crash if provided parameter is still *radius*.

```suggestion
          radius,
          diameter = radius || DEFAULT_DIAMETER,
```

Also, keep unit tests for both diameter AND radius params.

mabhub: I think you should keep backward compatibility by still allowing to have a *radius* parameter. This way, your PR won't be a breaking change anymore.

So you keep your changes about renaming and using *diameter*, but you can adjust so the app won't crash if provided parameter is still *radius*.

```suggestion
          radius,
          diameter = radius || DEFAULT_DIAMETER,
```

Also, keep unit tests for both diameter AND radius params.

PR #91, opened 2019-09-06
Fix display tooltip behavior

mabhub:
### What was wrong

When moving mouse from a map feature having a tooltip, to an element that was inside the map but not the map properly, the tooltip was shown.

Put another way: map feature tooltips do not disapear when mouse pointer is over map controls.

image
image

### What does this PR

- Add a fix to consider having mouse over map controls as not really beeing *over* the map.
- Add a fuse to avoid displaying a feature tooltip if mouse pointer is not over the map anymore.

### DIFF
--- a/src/modules/Map/InteractiveMap/InteractiveMap.js
+++ b/src/modules/Map/InteractiveMap/InteractiveMap.js
@@ -44,6 +44,8 @@ const generateTooltipContainer = ({ fetchProperties, properties, template, conte
 
 export const getHighlightLayerId = (layerId, type = 'fill') => `${type}-${layerId}-highlight`;
 
+const isChildOfAny = (target, elementsArray) => elementsArray.some(element => element.contains(target));
+
 const getUniqueLegends = legends => {
   const uniques = [];
   legends.forEach(legend => uniques.find(({ title, items }) =>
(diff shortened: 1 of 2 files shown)


src/modules/Map/InteractiveMap/InteractiveMap.js
```diff
@@ -44,6 +44,39 @@ const generateTooltipContainer = ({ fetchProperties, properties, template, conte
...
+
+  const mapControlGroups = mapContainer.querySelectorAll('.mapboxgl-ctrl-group');
+  const isPointerOverControl = isChildOfAny(target, [...mapControlGroups]);
+
+  // Event is from a Map control button (so, no directly from Map)
+  if (isPointerOverControl) return false;
+
+  return true;
+};
+
```
dtrucs: Could you test this function?

src/modules/Map/InteractiveMap/InteractiveMap.js
```diff
@@ -44,6 +44,39 @@ const generateTooltipContainer = ({ fetchProperties, properties, template, conte
 
 export const getHighlightLayerId = (layerId, type = 'fill') => `${type}-${layerId}-highlight`;
 
+const isChildOfAny = (target, elementsArray) =>
```
hadrienl: Useless function. It's used once and its length is only one line.

src/modules/Map/InteractiveMap/InteractiveMap.js
```diff
@@ -44,6 +44,39 @@ const generateTooltipContainer = ({ fetchProperties, properties, template, conte
...
+  const mapContainer = map.getContainer();
+  const isPointerOverContainer = mapContainer.contains(target);
+
+  // Event is not from any DOM element of Map
+  if (!isPointerOverContainer) return false;
+
+  // Event IS from an element of Map
+
+  const mapControlGroups = mapContainer.querySelectorAll('.mapboxgl-ctrl-group');
+  const isPointerOverControl = isChildOfAny(target, [...mapControlGroups]);
```
hadrienl:
```suggestion
  const isPointerOverControl = [...mapControlGroups].some(element => element.contains(target));
```

src/modules/Map/InteractiveMap/InteractiveMap.js
mabhub: hmm.. right. When I created it, it was use twice, and usefull for code readability. I agree that it's required anymore.

src/modules/Map/InteractiveMap/InteractiveMap.js
mabhub: It is already tested through use cases. There's just a lack concerning a branch (where `isPointerOverControl`is `false`) and I'm not really sure for now about how to create this case in existing tests.  
If you have any suggestions on how to produce the case, I'm open to suggestions.

mabhub: I resolve this conversation 'cause we wrote the tests together.

PR #92, opened 2019-09-06
Move search from project

hadrienl:


src/modules/Visualizer/services/layersTreeUtils.test.js
```diff
@@ -420,3 +448,86 @@ it('should sort custom layers', () => {
     id: '2',
   }]);
 });
+
+it('shoulod fetch property value', async () => {
```
SebCorbin: Small typo

src/modules/Visualizer/services/layersTreeUtils.js
```diff
@@ -186,6 +189,84 @@ export const sortCustomLayers = (customLayers, layersTree) => {
...
+      name: 'min',
+    }],
+    size: 0,
+  });
+  const { aggregations: { min: { value: min } = {}, max: { value: max } = {} } } = results;
+
+  return { min: min && Math.round(min), max: min && Math.round(max) };
+};
+
+export const layersTreeToStory = layersTree => {
```
SebCorbin: A bit of documentation here?

src/modules/Visualizer/services/search.js
```diff
@@ -0,0 +1,193 @@
...
+import bodybuilder from 'bodybuilder';
+import debounce from 'lodash.debounce';
+
+export const DEFAULT_INDEX = 'features';
+export const MAX_SIZE = 10000;
+export const SEARCHES_QUEUE = new Set();
+
+export const getExtentWithPadding = (map, { top, left, width, height }) => {
+  const cUL = map.unproject([left, top]).toArray();
+  const cBR = map.unproject([width + left, height + top]).toArray();
```
SebCorbin: Maybe use more explicit names ?

src/modules/Visualizer/services/search.js
```diff
@@ -0,0 +1,193 @@
...
+};
+
+export const buildQuery = ({
+  query = '',
+  boundingBox,
+  size = MAX_SIZE,
+  properties/* = { propName: value }, { propName: { value, type: 'term'} } */,
+  include,
+  exclude,
+  aggregations/* = [{ type, field, name, options }] */,
```
SebCorbin: Maybe document this? (parameters types, what they do, etc.)

src/modules/Visualizer/services/search.test.js
```diff
@@ -0,0 +1,441 @@
...
+            query: '*foo*',
+          },
+        },
+      },
+    },
+    size: 42,
+  });
+});
+
+it('should build query with query', () => {
```
SebCorbin: This outlines the fact that the parameter query may not be the best name

src/modules/Visualizer/services/layersTreeUtils.js
```diff
@@ -186,6 +189,84 @@ export const sortCustomLayers = (customLayers, layersTree) => {
...
+        size: MAX_SIZE,
+      },
+    }],
+    size: 0,
+  });
+  const { aggregations: { values: { buckets } } } = results;
+  return buckets.map(({ key }) => key);
+};
+
+export const fetchPropertyRange = async (layer, { property }) => {
```
SebCorbin: Is layer a layer object?

src/modules/Visualizer/services/layersTreeUtils.js
```diff
@@ -186,6 +189,84 @@ export const sortCustomLayers = (customLayers, layersTree) => {
   return newCustomLayers;
 };
 
+export const fetchPropertyValues = async (layer, { property }) => {
```
SebCorbin: Is layer a layer object?

src/modules/Visualizer/services/search.js
hadrienl: Changed ;)

src/modules/Visualizer/services/search.test.js
hadrienl: It's related to elastic search https://bodybuilder.js.org/docs/#query

SebCorbin: `should build elasticsearch query with query string`?

src/modules/Visualizer/services/layersTreeUtils.js
```diff
@@ -186,6 +189,97 @@ export const sortCustomLayers = (customLayers, layersTree) => {
   return newCustomLayers;
 };
 
+/**
+ * Fetch property range
```
SebCorbin: Copy-paste fail?

src/modules/Visualizer/services/layersTreeUtils.js
```diff
@@ -186,6 +189,97 @@ export const sortCustomLayers = (customLayers, layersTree) => {
...
+      },
+    }],
+    size: 0,
+  });
+  const { aggregations: { values: { buckets } } } = results;
+  return buckets.map(({ key }) => key);
+};
+
+/**
+ * Fetch property range
```
SebCorbin: Shouldn't it be more explicit? Like `Fetch min and max of property values`.

It's also missing the return type (which could mention the rounding of min/max)

src/modules/Visualizer/services/layersTreeUtils.js
SebCorbin: Is it useful to export this function if only used in stories? It could be moved to the stories directories, thus not being compiled to production bundle

hadrienl: For now, the Story component is not migrated here. We may move it in the future.

SebCorbin: Oh ok, `Story` component != story from storybook, that a bit confusing

PR #95, opened 2019-09-09
Hooks are so overrated

SebCorbin:


src/modules/Forms/Controls/RangeSlider.js
```diff
@@ -1,38 +1,28 @@
...
-      {...props}
-    />
-  );
-};
+    return (
+      <BPRangeSlider
+        className="control--range__slider"
+        value={range}
+        onRelease={onChange}
+        onChange={this.setRange}
```
mabhub: Where does the `setRange` method come from?

src/modules/Forms/Controls/RangeSlider.js
```diff
@@ -1,38 +1,28 @@
-import React, { useState, useEffect } from 'react';
+import React  from 'react';
```
mabhub: There's an extraneous whitespace between `React` and `from`.

```suggestion
import React from 'react';
```

src/modules/Forms/Controls/RangeSlider.js
SebCorbin: From the super-class `RangeComponent`, I put all common logic into it

mabhub: Oh right! Sorry for the mis, I did not see that `RangeSlide` is extending `RangeComponent` instead of `React.Component`.

PR #96, opened 2019-09-09
Change button tooltip behavior with touch device

MarianneLa:


hadrienl:
I would have put withDevice in `components` folder. It's a hoc, but it's also a component.

src/modules/Map/Map/components/SearchControl/__snapshots__/SearchControl.test.js.snap
```diff
@@ -2,45 +2,32 @@
 
 exports[`should render 1`] = `
 Array [
-  <span
-    className="bp3-popover-wrapper"
+  <button
```
mabhub: Also, check why all these snapshots have change when they shouldn't.

PR #99, opened 2019-09-10
Convert Checkbox to functional component

SebCorbin:


src/modules/Forms/Controls/Checkboxes/Checkbox.js
```diff
@@ -2,31 +2,29 @@ import React from 'react';
...
-    onToggle: PropTypes.func,
-  };
-
-  static defaultProps = {
-    onToggle () {},
-    checked: false,
-    value: '',
-    label: '',
-  };
+function Checkbox (props) {
```
mabhub:
```suggestion
const Checkbox = ({ value, label, checked, onToggle }) => {
```

PR #100, opened 2019-09-10
Move Details from project

hadrienl:


src/modules/Visualizer/Details/Details.js
```diff
@@ -0,0 +1,179 @@
...
+    if (prevId !== _id) {
+      this.updateIndex(_id);
+    }
+
+    if (visible !== prevVisible) {
+      this.switchVisibility(visible);
+    }
+  }
+
+  getIndexFeature = index => {
```
SebCorbin:
```suggestion
  getFeatureIndex = index => {
```

SebCorbin: Or `getIndex`

hadrienl: My goal is not to refacto the existant codebase, but only to migrate it and add missing tests.

PR #102, opened 2019-09-11
Move MapNavigation, PrivateLayers and TooManyResults

hadrienl:


src/modules/Visualizer/Story/index.js
```diff
@@ -0,0 +1,5 @@
+import Story from './Story';
+//import { connectView } from '../context';
```
mabhub: ❌ Extraneous comment

src/modules/Visualizer/Story/index.js
```diff
@@ -0,0 +1,5 @@
...
+
+//export default connectView('map', 'setLegends')(Story);
```
mabhub: ❌ Extraneous comment

src/modules/Visualizer/Story/index.js
hadrienl: It's temporary. I'll remove this comments when connectView will be migrated.

PR #107, opened 2019-09-13
Remove tooltip interactive map mobile

MarianneLa:


src/modules/Map/InteractiveMap/InteractiveMap.js
```diff
@@ -367,10 +369,11 @@ export class InteractiveMap extends React.Component {
     element,
     className,
   }) => {
-    const { history } = this.props;
+    const { history, isMobileSized } = this.props;
     const { map, isPointerOverMap } = this;
 
     if (isPointerOverMap === false) return;
+    if (isMobileSized) return;
```
dtrucs: The `isMobileSized` boolean only defines the size of the screen. I'm not sure it's the best way to detect if the client has a touchscreen display or not.

In this case, i would remove the HOC and test the following condition instead: 
`if(!global.matchMedia("(hover: hover)").matches) return`

src/modules/Map/InteractiveMap/InteractiveMap.test.js
```diff
@@ -865,6 +865,22 @@ describe('Interactions', () => {
...
+    });
+    instance.setState = () => null;
+    instance.componentDidMount();
+    // instance.displayTooltip({
+    //   layerId: 'foo',
+    //   features: [{ properties: {} }],
+    //   event: { lngLat: { lng: 3, lat: 4 } },
+    //   template: 'bar',
+    // });
+
```
dtrucs: Maybe this test is irrelevant

src/modules/Map/InteractiveMap/InteractiveMap.js
```diff
@@ -661,4 +663,4 @@ export class InteractiveMap extends React.Component {
   }
 }
 
-export default connectState('initialState')(InteractiveMap);
+export default connectState('initialState')((InteractiveMap));
```
mabhub: There's an extraneous pair of parens.

src/modules/Map/InteractiveMap/InteractiveMap.js
```diff
@@ -133,6 +132,7 @@ export class InteractiveMap extends React.Component {
     })),
     onInit: PropTypes.func,
     onStyleChange: PropTypes.func,
+    isMobileSized: PropTypes.bool,
```
dtrucs: This propTypes is defined but never called after

src/modules/Map/InteractiveMap/InteractiveMap.js
```diff
@@ -141,6 +141,7 @@ export class InteractiveMap extends React.Component {
     legends: [],
     onInit () {},
     onStyleChange () {},
+    isMobileSized: false,
```
dtrucs: This defaultProps is setted but never called after

locales/fr/translation.json
```diff
@@ -144,8 +145,8 @@
         }
       },
       "too_many_results": "Votre recherche contient trop de résultats, veuillez affiner votre requête ou zoomer sur une zone plus restreinte.",
-      "fold": "Relier",
+      "too_many_results_plural": "",
```
dtrucs: Maybe this line is useless?

MarianneLa: done

src/modules/Map/InteractiveMap/InteractiveMap.js
MarianneLa: done

src/modules/Map/InteractiveMap/InteractiveMap.js
MarianneLa: done

src/modules/Map/InteractiveMap/InteractiveMap.js
MarianneLa: done

PR #114, opened 2019-09-26
Improve filters fetching performance and UX

### LINKED ISSUE #109: Values fetching could be improved
`LayerFetchValues` handle the fetching of property values through two functions in `LayersTreeProvider`: `fetchPropertyValues` and `fetchPropertyRange`.

These two do a lot of `resetState`(one before fetching, one after result), for each property, even if there is of request sent. This results in a lot of commit for React, and eventually a bad UX.

I suggest grouping the initial `resetState` for all properties, we could also implement a `loading` state. And grouping the result of the fetch so that the state is updated once (after all properties are done loading).

SebCorbin:
Fix #109 #88

### DIFF
--- a/src/modules/Forms/Controls/Checkboxes/Checkboxes.js
+++ b/src/modules/Forms/Controls/Checkboxes/Checkboxes.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { Spinner } from '@blueprintjs/core';
 
 import Checkbox from './Checkbox';
 import formatValues from '../formatValues';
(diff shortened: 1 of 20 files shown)


mabhub:
Could you do:
```jsx
{!!loading && <Spinner size={20} />}

{!loading && (
  // ...
)}
```
instead of:
```jsx
{loading ? <Spinner size={20} /> : (
  // ...
)}
```

?

I think multi-line ternary with JSX should be avoid as soon as any argument of the ternary is more than 2 ou 3 line long. Don't you?

mabhub:
Please add `loading` in `propTypes` _(at least a defaultProps)_

PR #115, opened 2019-09-27
Fix warning about initialContent and intialprops in Filter

DarkVisenya:


src/modules/Forms/Filters/Filters.js
```diff
@@ -10,7 +9,7 @@ import Switch from '../Controls/Switch';
 import MultiSelect from '../Controls/MultiSelect';
 import DateRangeInput from '../Controls/DateRangeInput';
 import Control from './Control';
-import translateMock from '../../../utils/translate';
+import translateTerra from '../../../utils/translate';
```
SebCorbin: Why change this name?

src/modules/Forms/Filters/Control.js
```diff
@@ -50,7 +50,7 @@ export class Control extends React.Component {
 
     return (
       <Component
-        translate={translate === defaultTranslate ? undefined : translate}
```
SebCorbin: Please check that this works in storybook before removing this

src/modules/Forms/Filters/Control.js
```diff
@@ -50,7 +50,7 @@ export class Control extends React.Component {
...
+        translate={translate}
```
mabhub: If you don't transform `translate` anymore, just let it inside spread props:

```jsx
  render () {
    const { component: Component, ...props } = this.props;
    const { values } = this.state;

    return (
      <Component
        {...props}
        values={values}
      />
    );
  }

```

src/modules/Forms/Filters/Control.js
DarkVisenya: I already test on storybook. It's why I had to mock translate in Select and Filters.

src/modules/Forms/Filters/Filters.js
DarkVisenya: An error, sorry

src/modules/Forms/Filters/Control.js
DarkVisenya: Done

PR #117, opened 2019-10-02
Increase device pixel ratio for printing

SebCorbin:


src/modules/Map/Map/components/PrintControl/export.js
```diff
@@ -32,10 +38,18 @@ export default async function exportPdf (map, orientation, format = 'a4') {
     ignoreElements: ({ className: classes }) => typeof classes === 'string' && classes.includes('mapboxgl-control-container'),
   });
 
-  // JsPdf will automatically detect and put the image full page
-  doc.addImage(renderedContainer, 'PNG', 0, 0);
+  // Put the image full page
+  const conversionFactor = 96 / 25.4;
```
dtrucs: What do these numbers mean?

SebCorbin: From pixels to mm

dtrucs: Sorry for the incomprehension, it was a suggestion to define these numbers with explicit const.

But after another review, i wonder where the `dpi / 96` computation is optimized for print? Is it a rule or have we passed a resolution high enough to say that everything is fine?

jrmi: I suggest that you add a comment to explain the meaning of this values and don't go further for now as this feature need a longer thinking, if you see what I mean. It's ok for you @dtrucs  ?

PR #118, opened 2019-10-04
Fix scss on hover in mapbox-crtl

DarkVisenya:


src/modules/Map/Map/Map.scss
```diff
@@ -34,15 +36,39 @@
...
+      background-color: #435461;
+    }
+  }
+
+  .mapboxgl-ctrl-background-styles {
+    button:hover {
+      background-color: #435461;
+    }
+  }
+
```
dtrucs: 1/ I think it's better to double the selector with the `:focus` pseudo-selector for navigation by keyboard purpose.

2/ You can DRY with comma it all the beginning of selectors:
```css
&.mapboxgl-ctrl-home,
&.mapboxgl-ctrl-print,
&.mapboxgl-ctrl-share,
.mapboxgl-ctrl-background-styles {
    button:hover,
    button:focus {
      background-color: #435461;
    }
}

```
Maybe is it possible to refacto these lines with a more generic selector (without suffix) like ?
```css
  &.mapboxgl-ctrl {
    button:hover,
    button:focus {
      background-color: #435461;
    }
  }
```

DarkVisenya: I can't use directly .mapboxgl-ctrl so I applied your first suggestion

PR #119, opened 2019-10-10
Fix share control popover gap on mobile

MarianneLa:


src/modules/Map/Map/components/ShareControl/ShareControl.test.js
```diff
@@ -49,24 +52,27 @@ it('should display correct url with map named', () => {
...
+  // const selectionRange = jest.fn();
   const instance = wrapper.instance();
-
-  wrapper.find('.bp3-input').props().onClick({
-    target: {
-      setSelectionRange: selectionRange,
-      value: 'coucou',
-    },
-  });
+  // console.log(wrapper.find('.bp3-input') ? 'youpi' : 'nope');
```
DarkVisenya: Why not delete these lines ?

PR #125, opened 2019-10-22
Filter all features when there are nothing to display

dtrucs:


src/modules/Visualizer/services/layersTreeUtils.js
```diff
@@ -129,9 +129,13 @@ export const filterFeatures = (
     if (!active || !layers) return;
     const layerFeatures = features
       .find(({ layer: fLayer }) => fLayer === layer);
-    const { features: ids = [] } = layerFeatures || {};
+    const { features: ids } = layerFeatures || {};
+
+    let filter = null;
```
mabhub:
```js
const filterFromFeatureIds = ids => {
  if (!Array.isArray(ids)) {
    return null;
  }

  if (!ids.length) {
    return false;
  }

  return ['match', ['get', '_id'], ids, true, false];
};

// -----------------------------------------------------


const filter = filterFromFeatureIds(ids);

```

PR #126, opened 2019-10-22
Do not manage min/max limits in rangeNumeric

DarkVisenya:


DarkVisenya:
> Please explain _(at least, in PR description, but an explicit title whould also improve next changelog)_ what change(s) is made to RangeNumeric component logic
> 
> Although, I clearly see in source that min/max of Range component are not taken in account anymore, it does not explicitly says if these min/max are… not needed? useless? meaningless? done somewhere else? or even if they were causing a bug.

Currently, the limits of rangeNumeric widget are too restrictive. User cannot enter values below/higher the range limits (which corresponds to the min and max in the props).

> Example : if min = 12, the value of the rangeNumeric input cannot be 0. 

So, to avoid user frustration and to correct the interface problem, we decide to remove this behavior and allow the user to enter wished value.

src/modules/Forms/Controls/RangeNumeric.test.js
```diff
@@ -42,3 +42,14 @@ it('should handle danger intent', () => {
   expect(wrapper.find('NumericInput').first().props().intent).toEqual('danger');
   expect(onChange).not.toHaveBeenCalled();
 });
+
+it('should not take consideration of props min or max', () => {
```
SebCorbin: You should have something like 
```
 wrapper.find('NumericInput').first().prop('onValueChange')(0);
  expect(onChange).toHaveBeenCalledWith([0, 20]);
```

DarkVisenya: Done

PR #133, opened 2019-11-06
Use native hash option

SebCorbin:


mabhub:
hurray ! :confetti_ball:

PR #134, opened 2019-11-13
Drop expired or invalid jwt token

mabhub:


src/modules/Auth/services/auth.js
```diff
@@ -21,53 +52,43 @@ export async function obtainToken (email, password) {
...
 export function getToken () {
-  return global.localStorage.getItem(TOKEN_KEY);
+  const storedToken = global.localStorage.getItem(TOKEN_KEY);
+
+  return checkToken(storedToken)
+    ? storedToken
+    : undefined;
+}
+
+export function invalidToken () {
```
jrmi:
```suggestion
export function clearToken () {
```
?

src/modules/Auth/services/auth.test.js
```diff
@@ -1,41 +1,50 @@
+import b64u from 'base64url';
```
dtrucs: Can't we use `global.btoa` / `atob` to encode/decode base64 string?
Actually, i think it's sad to import a package just for test cases.

mabhub: Base64url (RFC 4648) is not exactly the same as simple Base64 (RFC 3548).
It is Base64 with some transformations for beeing url safe. JWT spec is using Base64url.

[base64url package](https://github.com/brianloveswords/base64url/blob/master/src/base64url.ts) does Base64 encoding/decoding with these *url safe* transformation management.

This implementation should be used preferably rather than atob/btoa.

PR #140, opened 2019-11-20
Improve LayersTreeItem performance

SebCorbin:


src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeItem.js
```diff
@@ -18,6 +19,8 @@ import { displayWarningAccordingToZoom } from '../../services/warningZoom';
 import LayersTreeExclusiveItemsList from './LayersTreeExclusiveItemsList';
 import Tooltip from '../../../../components/Tooltip';
 
+const getHtmlId = moize(layer => btoa(JSON.stringify(layer).replace(/\W/g, '')));
```
SebCorbin: Use an uuid instead ?

src/modules/Visualizer/LayersTree/LayersTreeItem/WarningZoom/index.js
```diff
@@ -1,4 +1,17 @@
...
+  if (!isActive) return children;
+  const { showWarning, minZoomLayer } = processWarningAccordingToZoom(map, layer);
+  return (
+    <WrappedComponent display={showWarning} minZoomLayer={minZoomLayer} {...props}>
+      {children}
+    </WrappedComponent>
+  );
+};
+
+export default withWarningAccordingToZoom(connectLayersTree('translate')(WarningZoom));
```
mabhub: As soon as there is more than one HOC, it's a good idea to use a `compose()` function.

`compose()` may comes from a local helper like [the one used in terra-admin](https://github.com/Terralego/terra-admin/blob/master/src/utils/compose.js) or from any package as a dependency _(attention: all compose functions do not have the same signature. Sometimes arguments order and/or initial argument differ)_.

```suggestion
export default compose(
  withWarningAccordingToZoom,
  connectLayersTree('translate'),
)(WarningZoom);
```

src/modules/Visualizer/services/warningZoom.js
```diff
@@ -18,17 +16,22 @@ export function displayWarningAccordingToZoom (map, layer) {
...
+export default function processWarningAccordingToZoom (map, layer) {
+  if (!map) return {};
+
+  const currentZoom = map.getZoom();
+  const layerZoom = getLayerZoom(layer, map);
   const isNotDisplayed = currentZoom < layerZoom.minzoom || currentZoom > layerZoom.maxzoom;
 
-  return { display: isNotDisplayed, minZoomLayer: layerZoom.minzoom };
+  return { showWarning: isNotDisplayed, minZoomLayer: layerZoom.minzoom };
 }
```
mabhub: Always prefer arrow function.

_Function expression (`const pouet = function () {}`)_ and _named function (`function pouet () {}`)_ should be kept for specific cases like class/object methods (who should provide a `this` context).
To sum up: when there's no `this` inside, it should probably be an arrow function.

```js
const processWarningAccordingToZoom = (map, layer) => {
  // ...
};

export default processWarningAccordingToZoom;
```

or _(but less explicit)_
```js
export default (map, layer) => {
  // ...
};
```

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeItem.test.js
```diff
@@ -64,8 +66,8 @@ jest.mock('./FiltersPanel', () => function FiltersPanel () {
 });
 
 jest.mock('../../services/warningZoom', () => ({
-  displayWarningAccordingToZoom (map) {
-    return { minZoomLayer: 14, display: map && true };
+  processWarningAccordingToZoom (map) {
+    return { minZoomLayer: 14, showWarning: map && true };
```
dtrucs: `&& true`?
Maybe `!!map`is better

SebCorbin: Haha yes, that's a leftover

PR #141, opened 2019-11-21
Drop state in MultiSelect component

SebCorbin:
Huge performance gap on this by drop the query management (which is handled internally by BPMultiSelect)

Another PR will be done for the Select component too, if applicable

### DIFF
--- a/src/modules/Forms/Controls/MultiSelect.js
+++ b/src/modules/Forms/Controls/MultiSelect.js
@@ -6,6 +6,7 @@ import classnames from 'classnames';
 import {
   Button,
   Intent,
+  Menu,
   MenuItem,
   Spinner,
 } from '@blueprintjs/core';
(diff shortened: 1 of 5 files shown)


src/modules/Forms/Controls/MultiSelect.js
```diff
@@ -14,6 +15,23 @@ import { MultiSelect as BPMultiSelect } from '@blueprintjs/select';
...
 
+/**
+ * Highlight with <strong> element
+ *
+ * @param {string} item - Item to highlight
+ * @param {string} query - Query to use as highlight
+ * @param {RegExp} regExp - Regular expression to use as splitter (must use capturing parenthesis)
+ * @return {Array} Can be used as React.el
+ */
+const highlightItem = (item, query, regExp) =>
```
dtrucs: `query` parameter seems useless?

src/modules/Forms/Controls/MultiSelect.js
```diff
@@ -95,63 +90,94 @@ export class MultiSelect extends React.Component {
...
+              const regExp = new RegExp(`(${query})`, 'i');
               return (
                 <MenuItem
                   active={modifiers.active}
                   icon={value.includes(item) ? 'tick' : 'blank'}
                   key={item}
                   onClick={handleClick}
-                  text={item}
+                  text={query && highlightSearch
+                    ? <span>{highlightItem(item, query, regExp)}</span>
```
dtrucs: Maybe the `<span />` wrapper is useless

src/modules/Forms/Controls/MultiSelect.js
```diff
@@ -14,6 +15,23 @@ import { MultiSelect as BPMultiSelect } from '@blueprintjs/select';
...
+ */
+const highlightItem = (item, query, regExp) =>
+  // Complex highlighting because we must take case into account
+  item.split(regExp).reduce((accu, elem) =>
+    // Use a reducer for a JSX join
+    [...accu, accu.length % 2 ? (
+      // Then get the case-friendly query part of item
+      <strong key={accu.length}>{elem}</strong>
+    ) : elem], []);
+
```
dtrucs: The function cannot be made with a simple array map method ?
`item.split(regExp).map((elem, i) => i % 2 ? <strong key={accu.length}>{elem}</strong> : elem)`

PR #144, opened 2019-11-25
Improve light mode

dtrucs:
Improve lightMode for BackgroundStyles and Search map control

### DIFF
--- a/src/modules/Map/Map/components/SearchControl/styles.scss
+++ b/src/modules/Map/Map/components/SearchControl/styles.scss
@@ -3,6 +3,7 @@
 .mapboxgl-ctrl-search {
   position: relative;
   overflow: visible;
+  z-index: 1;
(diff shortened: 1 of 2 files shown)


src/modules/Map/Map/components/SearchControl/styles.scss
```diff
@@ -3,6 +3,7 @@
 .mapboxgl-ctrl-search {
   position: relative;
   overflow: visible;
+  z-index: 1;
```
dtrucs: This commit/line is a little out of scope.
But this is to avoid the following issue:
image
I would like your advice to approve this or not

SebCorbin: You should add a comment in the code, so that it doesn't get removed, but that's fine with me

PR #153, opened 2019-12-03
Set and forward columnWidth props to BPTable

dtrucs:


src/modules/Table/Table.js
```diff
@@ -108,15 +108,28 @@ export class Table extends React.Component {
...
+      locales: customLocales,
+      loading,
+      initialSort,
+      columnWidths,
+      ...rest
+    } = this.props;
     const locales = { ...DEFAULT_LOCALES, ...customLocales };
     const { filteredColumns, filteredData } = this;
 
+    const setColumnsWidths = columnWidths && columnWidths.length === filteredColumns.length
```
mabhub:
```suggestion
    const columnWidthsProp = columnWidths && columnWidths.length === filteredColumns.length
```

src/modules/Table/Table.js
```diff
@@ -108,15 +108,28 @@ export class Table extends React.Component {
...
+      ? { columnWidths }
+      : {};
+
     return (
       <div className="table">
         <Header title={title} columns={columns} onChange={this.onHeaderChange} locales={locales} />
         <TableComponent
-          {...this.props}
+          {...rest}
+          {...setColumnsWidths}
```
mabhub:
```suggestion
          {...columnWidthsProp}
```

src/modules/Table/Table.js
```diff
@@ -108,15 +108,28 @@ export class Table extends React.Component {
...
+    const setColumnsWidths = columnWidths && columnWidths.length === filteredColumns.length
+      ? { columnWidths }
+      : {};
+
     return (
       <div className="table">
         <Header title={title} columns={columns} onChange={this.onHeaderChange} locales={locales} />
         <TableComponent
-          {...this.props}
+          {...rest}
```
mabhub: I would have put `...rest` after all props to always allow overriding from parent component.

dtrucs: For this case, i think we need to keep at the top. It's only for forwarding extra props from parent component to BlueprintJS Table component.
Also, If we set `rest` after all props, it will override these two following props 
```js
columns={filteredColumns}
data={filteredData}
```

PR #157, opened 2019-12-09
Fix bug on layer selection

jrmi:


src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeExclusiveItemsList/Selector/Selector.js
```diff
@@ -66,7 +68,7 @@ export class Selector extends React.Component {
             className="selector__select"
             onChange={this.onChange(name)}
             values={values}
-            value={values.find(({ value }) => selectorKey[name] === value)}
+            value={values.find(({ value }) => (selectorKey ? selectorKey[name] === value : false))}
```
mabhub: Could you put the `values.find()` in a `const` just before the `return`?

jrmi: I've made a function instead as we need values and name params from the `.map`.

PR #159, opened 2019-12-11
Remove warning on duplicate key for layer groups

jrmi:


jrmi:
I've just created this related issue #161

PR #163, opened 2019-12-17
Improve insertions of layers

dtrucs:
When inserting/deleting layers from the map, the map keeps a certain order driven by the type of layer.
The order is defined by this following constante `LAYERS_TYPES = ['background', 'raster', 'hillshade', 'heatmap', 'fill', 'fill-extrusion', 'line', 'circle', 'symbol']`; <- Left at the bottom; Right at the top.

1/ Move certain functions from the Map component to MapUtils
2/ Keep a better order
For example, if we try to insert a layer with a `fill` type and there isn't other layer with this type, this layer will be positioned after **all layers**. With this PR, it would try to insert it after a `heatmap` type layer, if there is not `heatmap` type layer, it would try with `hillshade` type layer, etc.
3/ No consideration for all `draw map Layer`. These must stay on top and not influence the order of others

### DIFF
--- a/src/modules/Map/Map/Map.js
+++ b/src/modules/Map/Map/Map.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import 'mapbox-gl/dist/mapbox-gl.css';
 import { detailedDiff } from 'deep-object-diff';
 
+import { getOrderedLayers, getLayerBeforeId, LAYER_TYPES_ORDER } from '../services/mapUtils';
 import { updateCluster } from '../services/cluster';
 
 import SearchControl from './components/SearchControl';
@@ -43,20 +44,6 @@ export const DEFAULT_CONTROLS = [{
   control: CONTROL_SCALE,
   position: CONTROLS_BOTTOM_LEFT,
 }];
-
-export function getLayerBeforeId (type, layers) {
-  const sameTypes = layers.filter(({ type: lType }) => type === lType);
-
-  if (!sameTypes.length) return undefined;
-
-  const lastOfSameType = sameTypes[sameTypes.length - 1];
-
-  const pos = layers.findIndex(({ id }) => id === lastOfSameType.id) + 1;
-
-  return layers[pos] && layers[pos].id;
-}
-
-export const LAYERS_TYPES = ['background', 'raster', 'hillshade', 'heatmap', 'fill', 'fill-extrusion', 'line', 'circle', 'symbol'];
 
 const defaultTranslate = translateMock({
   'terralego.map.zoom_in_control.title': 'Zoom in',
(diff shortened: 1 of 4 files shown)


src/modules/Map/services/mapUtils.js
```diff
@@ -213,12 +214,48 @@ export function fitZoom ({ feature, map, padding = 0 }) {
   map.fitBounds(bbox({ type: 'FeatureCollection', features }), { padding });
 }
 
+const getTypeBefore = (type, types = LAYERS_TYPES) => {
+  const index = types.findIndex(t => t === type) - 1;
+  return index > -1 ? types[index] : undefined;
+};
+
+export const getLayerBeforeId = (type, layers) => {
```
SebCorbin: Could you document this? The name of this function does not really help understand its meaning

src/modules/Map/Map/Map.js
```diff
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import 'mapbox-gl/dist/mapbox-gl.css';
 import { detailedDiff } from 'deep-object-diff';
 
+import { getOrderedLayers, getLayerBeforeId, LAYERS_TYPES } from '../services/mapUtils';
```
SebCorbin: `LAYERS_TYPES` does not seem to be used

src/modules/Map/services/mapUtils.js
```diff
@@ -4,6 +4,7 @@ import moize from 'moize';
 import { PREFIXES } from './cluster';
 
 export const PREV_STATE = {};
+export const LAYERS_TYPES = ['background', 'raster', 'hillshade', 'heatmap', 'fill', 'fill-extrusion', 'line', 'circle', 'symbol'];
```
SebCorbin: `LAYER_TYPES` may be more accurate

mabhub: `TYPES_ORDER`?

SebCorbin: > `TYPES_ORDER`?

True, `LAYER_TYPES_ORDER` then ?

src/modules/Map/Map/Map.js
dtrucs: We use it in the prop types to make sure the item's type of the prop `layers` is an equivalent of the item of `LAYERS_TYPES`

src/modules/Map/services/mapUtils.js
dtrucs: Ok i'm going to modify it

src/modules/Map/services/mapUtils.js
dtrucs: I agree, i'm on it

src/modules/Map/services/mapUtils.js
dtrucs: I didn't see other comments above the mine. I'm not sure about `ORDER` because we also use it as a list of all possible types of layers

src/modules/Map/Map/Map.js
SebCorbin: Oh yes, sorry

src/modules/Map/services/mapUtils.js
dtrucs: Finally, i think it's fine with order. It will avoid to have the order modified in the future (for example, a developer who would strictly want to order as alphabetical sorting for readability)

PR #164, opened 2019-12-17
Add custom control to the list of Map controls

dtrucs:


src/modules/Map/Map/Map.js
```diff
@@ -469,6 +471,19 @@ export class MapComponent extends React.Component {
           map.addControl(controlInstance, position);
           break;
         }
+        case CONTROL_CUSTOM: {
```
jrmi: Can we have more than one custom control ?

dtrucs: For sure,
We can also add all other controls X times on the map.
Nowadays, in our projects, we compare the type of control to remove it from the map.
But it is also possible to pass all the parameters we want (like an ID) when creating a control and use them to compare and remove it.

PR #167, opened 2020-01-10
Handle indexes in Search client

SebCorbin:


src/modules/Visualizer/services/search.test.js
```diff
@@ -415,23 +441,29 @@ it('should search many queries', () => {
...
+        query: {
+          query_string: {
+            query: '*bar*',
+          },
         },
+        size: MAX_SIZE / 2,
       },
-      size: MAX_SIZE / 2,
-    }],
+    ],
```
SebCorbin: These are just easier to read

src/modules/Visualizer/services/search.js
```diff
@@ -164,9 +178,13 @@ export class Search {
     index = DEFAULT_INDEX,
   */}) {
     const body = buildQuery(query);
+    const { index = DEFAULT_INDEX } = query;
```
dtrucs: `= DEFAULT_INDEX` seems useless

src/modules/Visualizer/services/search.js
```diff
@@ -187,21 +212,35 @@ export class Search {
     if (!queries.length) throw new Error('`queries` must not be empty');
 
     const searches = queries
-      .map(({ size = MAX_SIZE / queries.length, ...query }) => buildQuery({ size, ...query }))
-      .reduce((body, query) => [...body, { }, query],
+      .map(
+        ({ size = MAX_SIZE / queries.length, index, ...query }) =>
```
dtrucs: I would declare the following operation `MAX_SIZE / queries.length` above, outside the loop

src/modules/Visualizer/services/search.js
SebCorbin: Yes, it's a remnant of the comment two lines above, should I delete it everywhere in this file?

src/modules/Visualizer/services/search.js
```diff
@@ -187,21 +212,35 @@ export class Search {
...
+   */
   batchSearch = debounce(async () => {
-    const [bodies, resolves] = Array
+    // Consume the queue and reduce it to headers, bodies and resolvers
+    const [bodies, resolves, headers] = Array
       .from(SEARCHES_QUEUE.values())
-      .reduce(([allBodies, allResolves], { body, resolve }) =>
-        [[...allBodies, body], [...allResolves, resolve]], [[], []]);
+      .reduce(
+        ([allBodies, allResolves, allHeaders], { body, resolve, header = {} }) =>
```
mabhub: For readability purpose, wouldn't this reduce function be best put in a named function?

src/modules/Visualizer/services/search.js
SebCorbin: Deleted

src/modules/Visualizer/services/search.js
SebCorbin: The new indentation makes it a bit easier to read, and I've made the order more logical (header -> body -> resolve)

src/modules/Visualizer/services/search.js
```diff
@@ -164,9 +174,13 @@ export class Search {
     index = DEFAULT_INDEX,
   */}) {
     const body = buildQuery(query);
+    const { index } = query;
     const action = {
       body,
     };
+    if (index) {
```
dtrucs: I don't know really what the code does, but just before there was a condition like `if index !== DEFAULT_INDEX`. Is it no longer relevant?


Sorry if i included a misunderstanding with ma last review (remove the default value of `index` 4 lines before.)

PR #168, opened 2020-01-10
Add pre-commit hooks

SebCorbin:


src/modules/Map/InteractiveMap/components/Legend/Legend.test.js
```diff
@@ -87,11 +87,11 @@ it('should render a manual legend', () => {
...
 
-<square color="red" /> foo  
-<square color="green" /> bar  
+<square color="red" /> foo
+<square color="green" /> bar
 
-<circle color="red" /> foo  
-<circle color="green" /> bar  
+<circle color="red" /> foo
+<circle color="green" /> bar
```
mabhub: EOL whitespaces should NOT be removed from markdown templates.

src/stories/modules/Map/Map.md
```diff
@@ -14,18 +14,17 @@ You can import the default set if you want just to add some more controls :
...
+It is defined by an object with `top`, `bottom`, `left` and `right` keys,
 which take number of pixels as values.
 
-`offset` attribute is an array to define the center of the given bounds relative to the map's center, 
+`offset` attribute is an array to define the center of the given bounds relative to the map's center,
 measured in pixels.
 
-You can change the default map bounds options at the time of first map initialisation or when the flyTo and  animations are running.  
-
+You can change the default map bounds options at the time of first map initialisation or when the flyTo and  animations are running.
```
mabhub: EOL whitespaces should NOT be removed from markdown files.

src/stories/modules/Map/MapControls.md
```diff
@@ -31,9 +31,9 @@ Map component comes with a "in map" search control. It will helps you to insert
...
-  group: 'group title',  
-  results: [{  
-    label: 'result item label',  
-    center: [5.4859932, 43.3271871]  
-  }]  
+  group: 'group title',
+  results: [{
+    label: 'result item label',
+    center: [5.4859932, 43.3271871]
+  }]
```
mabhub: EOL whitespaces should NOT be removed from markdown files, but in this case, the code block should start with ` ```js ` and then having no EOL whitespaces.

mabhub:
I'm personnaly not fan of git pre commit hooks for enforcing linting automagicaly. But as it's an "opt-in" feature, I see no problem of adding a common config file.

However, people having enabled such a setup should be double vigilant not commiting changes not direcly related to their current code.

SebCorbin:
> I'm personnaly not fan of git pre commit hooks for enforcing linting automagicaly. But as it's an "opt-in" feature, I see no problem of adding a common config file.

It's not "magical", the CI will enforce linting eventually, it just check even before pushing, reducing noise on github

src/modules/Map/InteractiveMap/components/Legend/Legend.test.js
SebCorbin: These double spaces are not relevant for the test, and I'm not comfortable making an exception just for this

mabhub: Replacing these *double space* by a `<br />` may do the job in this case...

jrmi:
Closed as too old and need update.

PR #170, opened 2020-01-17
Retrieve objects from the updated key to add/remove customStyles

dtrucs:


src/modules/Map/Map/Map.js
```diff
@@ -328,16 +328,40 @@ export class MapComponent extends React.Component {
...
+    } = getListDiffed(deleted, prevCustomStyle);
+    const {
+      sources: updatedSources,
+      layers: updatedLayers,
+    } = getListDiffed(updated, customStyle);
+    const {
+      sources: addedSources,
+      layers: addedLayers,
+    } = getListDiffed(added, customStyle);
+
```
mabhub: This may benefit from a bit of [currying](https://en.wikipedia.org/wiki/Currying).

PR #174, opened 2020-02-17
Fix initial state managment

jrmi:
- default active layers were lost on view change after loading
- some readability enhancement

### DIFF
--- a/src/modules/State/Hash/withHashState.js
+++ b/src/modules/State/Hash/withHashState.js
@@ -13,6 +13,11 @@ export const DEFAULT_OPTIONS = {
   parseBooleans: true,
 };
 
+
+const getCurrentHashState = () => parse(window.location.hash, DEFAULT_OPTIONS);
+
+const getCurrentHashString = state => `#${stringify(state, DEFAULT_OPTIONS)}`;
+
 /**
  * Decorator for getting an initialState from hash.
  */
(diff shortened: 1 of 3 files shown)


src/modules/Visualizer/LayersTree/LayersTreeProvider/LayersTreeProvider.js
```diff
@@ -56,6 +56,10 @@ export class LayersTreeProvider extends React.Component {
   }
 
   componentDidMount () {
+    const { setCurrentState } = this.props;
+    // Reset current tree state from hash
+    setCurrentState({ layers: undefined });
```
jrmi: This is the actual useful change

src/modules/State/Hash/withHashState.js
```diff
@@ -32,6 +32,12 @@ export const withHashState = WrappedComponent => {
 
     options = DEFAULT_OPTIONS;
 
+    state = {};
+
+    componentWillMount () {
```
dtrucs: `UNSAFE_componentWillMount()`

jrmi: I put the code in constructor instead and remove deprecated willMount lifecycle.

src/modules/State/Hash/withHashState.js
```diff
@@ -32,6 +32,13 @@ export const withHashState = WrappedComponent => {
 
     options = DEFAULT_OPTIONS;
 
+    state = {};
+
+    constructor () {
+      super();
+      this.onHashChange();
+    }
```
dtrucs: Instead of this constructor, isn't easier to define directly the state above ?
```js
state = {
  initialState: this.getCurrentHashState(),
}
```

jrmi: `this` is not working in this context, but I put the function outside of the component.

PR #175, opened 2020-02-18
Fix popup logout confirmation display

MarianneLa:


dtrucs:
Maybe you could make some tests?

PR #178, opened 2020-02-25
Add message when filter values are empty

MarianneLa:


src/modules/Forms/Controls/NoValues.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import './index.scss';
+
+export const NoValues = ({ label }) => (
+  <div className="control-container">
+    <p className="control-label">{label}</p>
+    <p className="control-label-noValues">Aucun choix disponible</p>
```
mabhub: What about translation?

src/modules/Forms/Filters/Filters.js
```diff
@@ -41,18 +42,25 @@ export function getComponent (type, values, display, format) {
...
+            translate={translate}
+            component={getComponent(type, values, display, format)}
+            key={property}
+            type={type}
+            values={values}
+            property={property}
+            onChange={value => onChange({ ...properties, [property]: value })}
+            value={properties[property]}
+          />
+        )
```
mabhub: Avoid ternary for multi-line JSX :

```jsx
<div>
  {filters.map(({ type, values, property, display, format, label, ...props }) => {
    // Early return if `value` is an empty array
    if (values && values.length === 0) {
      return <NoValues label={label} />;
    }

    return (
      <Control
        {...props}
        label={label}
        translate={translate}
        component={getComponent(type, values, display, format)}
        key={property}
        type={type}
        values={values}
        property={property}
        onChange={value => onChange({ ...properties, [property]: value })}
        value={properties[property]}
      />
    );
  })}
</div>
```

locales/fr/translation.json
```diff
@@ -64,6 +64,7 @@
           "select_placeholder_plural": "Saisissez au moins {{count}} caractères",
           "too_many": "Trop de résultats, affinez votre recherche"
         },
+        "noValues": "",
```
mabhub: Should really this be empty?

src/modules/Forms/Controls/Checkboxes/Checkboxes.js
```diff
@@ -34,6 +35,7 @@ export class Checkboxes extends React.Component {
       <div className="control-container">
         <p className="control-label">{mainLabel}</p>
         {!!loading && <Spinner size={20} />}
```
dtrucs: The loading's propType is a boolean, you don't need to cast to boolean with `!!`.

src/modules/Forms/Controls/Checkboxes/Checkboxes.js
```diff
@@ -34,6 +35,7 @@ export class Checkboxes extends React.Component {
...
+        {!loading && values && values.length === 0 && <NoValues />}
```
dtrucs: The values default propType is an empty array, you don't need to verify if it's defined => 
`{!loading && !values.length && <NoValues />}`

PR #179, opened 2020-03-06
Add ability to sort layer by weight

jrmi:


src/modules/Map/Map/withMap.js
```diff
@@ -5,6 +5,113 @@ import { addMapDebug } from '../helpers/mapDebug';
...
 import 'mapbox-gl/dist/mapbox-gl.css';
 
+export const DEFAULT_LAYER_TYPES_WEIGHT = {
+  background: 100,
+  raster: 200,
+  hillshade: 300,
+  fill: 400,
+  line: 500,
+  heatmap: 600,
+  'fill-extrusion': 700,
```
frodrigo: fill-extrusion at 1000

src/modules/Map/Map/withMap.js
```diff
@@ -5,6 +5,113 @@ import { addMapDebug } from '../helpers/mapDebug';
...
+  /**
+   * Initialize weight for every auto loaded layer from background style.
+   */
+  initBaseStyleWeight () {
+    let prevWeight = 0;
+    const { layers } = this.map.getStyle();
+
+    // Compute weigth map for each layer
+    layers.reduce((acc, { type, id }, index) => {
+      if (index < layers.length - 1) {
```
dtrucs: you should declare a constante something like `const layersLengthMinus1 = layers.length - 1` above the reduce to avoid useless calculation

src/modules/Map/Map/withMap.js
```diff
@@ -5,6 +5,113 @@ import { addMapDebug } from '../helpers/mapDebug';
...
+        const { type: nextType } = layers[index + 1];
+        // if the layer is beetween two layers with same weight, the layer get the same weight
+        if (this.typesWeigth[nextType] === prevWeight) {
+          acc[id] = prevWeight;
+          return acc;
+        }
```
dtrucs: No return after this if?

src/modules/Map/Map/withMap.js
```diff
@@ -5,6 +5,113 @@ import { addMapDebug } from '../helpers/mapDebug';
...
+      }
+      prevWeight = this.typesWeigth[type];
+      acc[id] = this.typesWeigth[type];
```
dtrucs: Maybe you should declare `this.typesWegith[type]` before attributing it

src/modules/Map/Map/withMap.js
```diff
@@ -5,6 +5,113 @@ import { addMapDebug } from '../helpers/mapDebug';
...
+      return acc;
+    }, this.weights);
+  }
```
dtrucs: I'm not even sure of the utility of the reduce.
As you don't really use the acc, I think a forEach could do the job (or a `for` loop for optimization purpose)

```js
    const layersLengthMinusOne = layers.length - 1;

    layers.forEach(({ type, id }, index) => {
      if (index < layersLengthMinusOne) {
        const { type: nextType } = layers[index + 1];
        // if the layer is beetween two layers with same weight, the layer get the same weight
        if (this.typesWeigth[nextType] === prevWeight) {
          layers[id] = prevWeight;
        }
      }
      const nextWeight = this.typesWeigth[type];
      prevWeight = nextWeight;
      layers[id] = nextWeight;
    });
```

src/modules/Map/Map/withMap.js
```diff
@@ -5,6 +5,113 @@ import { addMapDebug } from '../helpers/mapDebug';
...
+    }
+    return undefined;
+  }
+
+  /**
+   * Return props from this object if exists otherwise from original map object
+   * @param {*} obj instance of proxified object
+   * @param {*} prop prop being accessed
+   */
+  get (obj, prop) {
```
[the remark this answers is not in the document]
jrmi: This is a proxy, so this function is used by javascript himself.

src/modules/Map/Map/withMap.js
```diff
@@ -5,6 +5,113 @@ import { addMapDebug } from '../helpers/mapDebug';
...
+   */
+  getLayerIdWithBiggerWeight (layers, weight) {
+    const isDrawLayer = layer => layer.id.startsWith('gl-draw');
+    const validLayers = layers.filter(layer => !isDrawLayer(layer));
+
+    for (let i = 0; i < validLayers.length; i += 1) {
+      if (this.weights[validLayers[i].id] > weight) {
+        return validLayers[i].id;
+      }
+    }
```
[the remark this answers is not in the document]
jrmi: I propose instead to set this layer with the bigger weight + 10.

src/modules/Map/Map/withMap.js
jrmi: But, i use the acc, this is the "this.weight" ?

src/modules/Map/Map/withMap.js
jrmi: The return is the fall back after. There is always a return, if I'm not mistaken.

src/modules/Map/Map/withMap.js
jrmi: But indeed, this can be rewritten in foreach.

src/modules/Map/Map/withMap.js
dtrucs: A return is missing in the case of `index < layers.length - 1 &&  this.typesWeigth[nextType] !== prevWeight`

src/modules/Map/Map/withMap.js
dtrucs: Maybe i'm wrong, keep the reduce if it works properly

src/modules/Map/Map/withMap.js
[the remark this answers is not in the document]
dtrucs: I'm afraid of the case where users set a significant weight on a custom layer to be sure it's at the top of the stack; and after realizing that it's above drawLayers.

src/modules/Map/Map/withMap.js
jrmi: If think we don't need extra return anymore with new code ?

src/modules/Map/Map/withMap.js
jrmi: Done

src/modules/Map/Map/withMap.js
jrmi: Done

src/modules/Map/Map/withMap.js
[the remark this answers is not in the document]
jrmi: Can you look the new code ?

src/modules/Map/Map/withMap.js
```diff
@@ -5,6 +5,113 @@ import { addMapDebug } from '../helpers/mapDebug';
...
+  /**
+   * Replace mapbox addlayer to sort layer by weight. For each layer type,
+   * a weight is associated. This weight is used unless the layer has an height property.
+   * In this case the weight became the property. Allow user to sort layer as he wants to.
+   * @param {*} layer the layer to be added
+   * @param {*} beforeId if beforeId is set, the weight is ignored.
+   */
+  addLayer (layer, beforeId) {
+    const { type, id, weight: layerWeight } = layer;
+    const weight = layerWeight || this.typesWeigth[type];
```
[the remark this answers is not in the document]
jrmi: Done

src/modules/Map/Map/withMap.js
jrmi: done

src/modules/Map/Map/withMap.js
```diff
@@ -5,6 +5,129 @@ import { addMapDebug } from '../helpers/mapDebug';
...
+
+const isDrawLayer = layer => layer.id.startsWith('gl-draw');
+
+export class WeigthedMapProxy {
+  /**
+   * Constructor
+   * @param {*} param0
+   * @param {*} param0
+   * @param {*} param0
+   */
```
dtrucs: Okay 💃

src/modules/Map/Map/withMap.js
```diff
@@ -5,6 +5,129 @@ import { addMapDebug } from '../helpers/mapDebug';
...
+      return this[prop];
+    }
+    return obj[prop];
+  }
+
+  /**
+   * Returns max weigth from current weight list
+   */
+  getMaxWeight () {
+    return Math.max.apply(null, Object.values(this.weights));
```
dtrucs: Why not `Math.max(...Object.values(this.weights))`

jrmi: done

src/modules/Map/Map/withMap.js
jrmi: done

src/modules/Map/Map/withMap.js
```diff
@@ -5,6 +5,127 @@ import { addMapDebug } from '../helpers/mapDebug';
...
+   * @param {*} beforeId if beforeId is set, the weight is ignored.
+   */
+  addLayer (layer, beforeId) {
+    const { type, id, weight: layerWeight } = layer;
+    let weight =
+      layerWeight !== undefined ? layerWeight : this.typesWeigth[type];
+
+    if (isDrawLayer(layer)) { // If it's a draw layer,
+      weight = layerWeight !== undefined ? layerWeight : this.getMaxWeight();
+    }
```
dtrucs: With a getter we can avoid some useless assignation
```
getWeight (layerWeight) {
  if (layerWeight === undefined) {
     return layerWeight;
  }
  if (isDrawLayer(layer)) {
     return this.getMaxWeight();
   }
   return this.typesWeigth[type];
}
/// ...
const weight = getWeight(layerWeight);
```

jrmi: Done

PR #180, opened 2020-03-10
Fix filter init value

jrmi:
And enhance multi-select and select.

### DIFF
--- a/src/modules/Forms/Controls/MultiSelect.js
+++ b/src/modules/Forms/Controls/MultiSelect.js
@@ -15,13 +15,22 @@ import { MultiSelect as BPMultiSelect } from '@blueprintjs/select';
 import { preventEnterKeyPress } from '../../../utils/event';
 import translateMock from '../../../utils/translate';
 import NoValues from './NoValues';
+import formatValues from './formatValues';
 
 export class MultiSelect extends React.Component {
   static propTypes = {
     value: PropTypes.arrayOf(PropTypes.string),
     label: PropTypes.string,
     onChange: PropTypes.func,
-    values: PropTypes.arrayOf(PropTypes.string).isRequired,
+    values: PropTypes.arrayOf(
+      PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.shape({
+          label: PropTypes.string.isRequired,
+          value: PropTypes.any.isRequired,
+        }),
+      ]),
+    ).isRequired,
     isSubmissionPrevented: PropTypes.bool,
     translate: PropTypes.func,
     loading: PropTypes.bool,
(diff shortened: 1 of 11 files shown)


src/modules/Forms/Controls/MultiSelect.js
```diff
@@ -46,15 +55,40 @@ export class MultiSelect extends React.Component {
     highlightSearch: true,
   };
 
+  state = {
+    values: [],
+    value: [],
+  }
+
+  static getDerivedStateFromProps ({ values, value }) {
+    if (!values) return null;
```
dtrucs: Refering to PropTypes, `values` is a required Array; the test should'nt be `if(!values.length) {`?

src/modules/Forms/Controls/MultiSelect.js
```diff
@@ -79,6 +113,7 @@ export class MultiSelect extends React.Component {
       ...props
     } = this.props;
 
+    const { values = [], value = [] } = this.state;
```
dtrucs: Default empty array values seems useless because these are already initialized in state object

src/modules/Forms/Controls/Select/Select.js
```diff
@@ -74,10 +76,11 @@ export class Select extends React.Component {
...
     } = this.props;
 
-    const { values } = this.state;
+    const { values = [] } = this.state;
     const { handleChange } = this;
-    const displayedValue = typeof value === 'object' ? value.label : value;
-    const rawValue = typeof value === 'object' ? value.value : value;
+
+    const result = values.find(({ value: itemValue }) => value === itemValue) || values[0];
+    const displayedValue = result ? result.label : '';
```
dtrucs: const { label: displayedValue = '' } = result;

src/modules/Forms/Controls/MultiSelect.js
jrmi: The test is then useless, i removed it.

src/modules/Forms/Controls/Select/Select.js
jrmi: Not working if result is undefined.

src/modules/Forms/Controls/MultiSelect.js
jrmi: done

src/modules/Forms/Controls/MultiSelect.js
jrmi: done

src/modules/Forms/Controls/Select/Select.js
dtrucs: Correct 
`const { label: displayedValue = '' } = result || {};`

jrmi: I keep the previous version.

PR #182, opened 2020-03-13
Handle auto generated legend

jrmi:
When we don't have label but boundaries.

### DIFF
--- a/src/modules/Map/InteractiveMap/components/Legend/Legend.js
+++ b/src/modules/Map/InteractiveMap/components/Legend/Legend.js
@@ -15,7 +15,6 @@ const computeLabel = (translate, label, boundaries, index, itemCount, rounding =
   if (label && label !== '') {
     return label;
   }
-
   let { lower: { value: lower }, upper: { value: upper } = {} } = boundaries;
   if (typeof lower === 'number') {
     if (typeof upper === 'number') { // We have two value
@@ -44,8 +43,8 @@ export const Legend = ({
   stackedCircles,
   translate,
 }) => {
-  const biggestDiameter = items.reduce((int, item) =>
-    Math.max(int, item.diameter || DEFAULT_DIAMETER), 0);
+  const biggestDiameter = items.reduce((int, { diameter = DEFAULT_DIAMETER }) =>
+    Math.max(int, diameter), 0);
 
   return (
     <div className={`tf-legend tf-legend--level-${level}`}>
@@ -80,14 +79,6 @@ export const Legend = ({
           boundaries,
           diameter = radius || DEFAULT_DIAMETER,
         }, index) => {
-          const computedLabel = computeLabel(
-            translate,
-            label,
-            boundaries,
-            index,
-            items.length,
-            rounding,
-          );
           if (subItems) {
             return (
               <Legend
@@ -110,6 +101,15 @@ export const Legend = ({
               />
             );
           }
+
+          const computedLabel = computeLabel(
+            translate,
+            label,
+            boundaries,
+            index,
+            items.length,
+            rounding,
+          );
 
           return (
             <div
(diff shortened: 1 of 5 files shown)


src/modules/Map/InteractiveMap/components/Legend/Legend.js
```diff
@@ -3,13 +3,34 @@ import PropTypes from 'prop-types';
...
 
+const computeLabel = (translate, label, boundaries, index, itemCount, rounding = null) => {
+  if (label && label !== '') {
+    return label;
+  }
+  let { lower: { value: lower }, upper: { value: upper } = {} } = boundaries;
+  if (typeof lower === 'number') {
+    if (typeof upper === 'number') { // We have two value
+      if (rounding !== null) {
+        lower = parseFloat(lower.toFixed(rounding)).toLocaleString();
```
dtrucs: `toLocaleString`?
Is it a date manipulation?
If not, `toString()` method should be more appropriate.

dtrucs: I'm not sure of that piece of code.
toFixed method should render a string value; you parseFloat it and then you restringify the whole thing.
Just `lower.toFixed(rounding)` isn't enough?

src/modules/Map/InteractiveMap/components/Legend/Legend.js
```diff
@@ -18,7 +39,9 @@ export const Legend = ({
   position,
   content,
   history,
+  rounding,
   stackedCircles,
+  translate,
 }) => {
   const biggestDiameter = items.reduce((int, item) =>
     Math.max(int, item.diameter || DEFAULT_DIAMETER), 0);
```
dtrucs: If item.diameter equals to `0`, is it ok to have `DEFAULT_DIAMETER` value as fallback?

dtrucs: Maybe
`const biggestDiameter = items.reduce((int, { diameter = DEFAULT_DIAMETER }) => Math.max(int, diameter), 0);`

src/modules/Map/InteractiveMap/components/Legend/Legend.js
```diff
@@ -3,13 +3,35 @@ import PropTypes from 'prop-types';
...
+import translateMock from '../../../../../utils/translate';
 import Template from '../../../../Template/Template';
 import CustomComponents from './CustomComponents';
 import Circle from './components/Circle';
 import Rect from './components/Rect';
 
 const DEFAULT_DIAMETER = 16;
 
+const computeLabel = (translate, label, boundaries, index, itemCount, rounding = null) => {
+  if (label && label !== '') {
```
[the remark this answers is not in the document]
jrmi: we don't want this test to be true if label is undefined.

src/modules/Map/InteractiveMap/components/Legend/Legend.js
jrmi: For first comment, not working only with date:
image

For second comment:

src/modules/Map/InteractiveMap/components/Legend/Legend.js
jrmi: Done

src/modules/Map/InteractiveMap/components/Legend/Legend.js
dtrucs: My bad.
I knew https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Date/toLocaleString and not https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString.

👍

src/modules/Map/InteractiveMap/components/Legend/Legend.js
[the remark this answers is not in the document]
dtrucs: But if the value equals to an empty string (= falsy), the code will return before the second test.

why just not if(!label)?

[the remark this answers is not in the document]
dtrucs: oulala, forget it

PR #186, opened 2020-03-29
Optimize table render

jrmi:
and many minor fixes.

### DIFF
--- a/src/hoc/withDeviceSize.js
+++ b/src/hoc/withDeviceSize.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import debounce from 'lodash.debounce';
 
 
 const DESKTOP_BREAKPOINT = 1025;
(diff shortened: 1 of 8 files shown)


src/hoc/withDeviceSize.js
```diff
@@ -9,28 +10,48 @@ export const withDeviceSize = ({
...
+      const width = global.innerWidth;
+      const newIsMobileSized = width < desktop;
+      const newIsPhoneSized = width <= phone;
+
+      if (newIsMobileSized !== isMobileSized) {
+        this.setState({ isMobileSized: newIsMobileSized });
+      }
+      if (newIsPhoneSized !== isPhoneSized) {
+        this.setState({ isPhoneSized: newIsPhoneSized });
+      }
```
dtrucs: Maybe, it's better to compare the state with `setState` callback instead of `this.state` which could be not updated.
```js
this.setState(({ isMobileSized, isPhoneSized }) => {
  return {
    isMobileSized: newIsMobileSized 
    isPhoneSized: newIsPhoneSized;
  }
})
```

jrmi: Can you check what I've done ?

src/modules/Forms/Controls/RangeNumeric.js
```diff
@@ -1,45 +1,69 @@
...
 import RangeComponent from './RangeComponent';
 
+const sanitizeProps = (({ fetchValues,
+  labelStepSize,
+  min,
+  max,
+  type,
+  onChange,
+  ...props }) => ({ ...props }));
+
```
mabhub:
```suggestion
const sanitizeProps = ({
  fetchValues,
  labelStepSize,
  min,
  max,
  type,
  onChange,
  ...props
}) => props;
```
or

```suggestion
const sanitizeProps = ({ fetchValues, labelStepSize, min, max, type, onChange, ...props }) => props;
```

src/modules/Forms/Controls/RangeNumeric.js
```diff
@@ -1,45 +1,69 @@
...
+      this.setState({ intentUp: Intent.DANGER });
+    }
+
+    if (lowerB > upperB) {
+      this.setState({ intentUp: Intent.DANGER, intentLow: Intent.DANGER });
+    }
+
+    if (lowerB <= upperB && lowerB >= min && upperB <= max) {
+      if (lowerB === min && upperB === max) {
+        this.setState({ intentLow: undefined, intentUp: undefined });
```
dtrucs: You can put this line outside the if/else (and therefore delete the second `setState`)

src/modules/Forms/Controls/RangeNumeric.js
```diff
@@ -1,45 +1,69 @@
...
+  onChangeDebounced = debounce(range => this.props.onChange(range), 300);
+
   onNumericInputChange = pos => bound => {
-    const { onChange } = this.props;
+    const { min, max } = this.props;
     const { range } = this.state;
+
     if (Number.isNaN(+bound)) return; // Sanity check that allows typing '-'
     range[pos] = bound;
     this.setState({ range });
```
dtrucs:
```suggestion
    this.setState(({ range: prevRange }) => {
      const nextRange = [...prevRange];
      nextRange[pos] = bound;
      return { range: nextRange }
    })
```

src/modules/Forms/Controls/RangeNumeric.js
jrmi: done

src/modules/Forms/Controls/RangeNumeric.js
jrmi: done

src/modules/Forms/Controls/RangeNumeric.js
jrmi: done

PR #188, opened 2020-04-02
Fix legend when no label and add Line symbol

jrmi:
- Legend was broken when there is no configured label.
- Line legend can now be added.

### DIFF
--- a/src/modules/Map/InteractiveMap/components/Legend/Legend.js
+++ b/src/modules/Map/InteractiveMap/components/Legend/Legend.js
@@ -11,11 +11,13 @@ import Rect from './components/Rect';
 
 const DEFAULT_DIAMETER = 16;
 
-const computeLabel = (translate, label, boundaries, index, itemCount, rounding = null) => {
+const computeLabel = (translate, label, boundaries = {}, rounding = null) => {
   if (label && label !== '') {
     return label;
   }
-  let { lower: { value: lower }, upper: { value: upper } = {} } = boundaries;
+
+  let { lower: { value: lower } = {}, upper: { value: upper } = {} } = boundaries;
+
   if (typeof lower === 'number') {
     if (typeof upper === 'number') { // We have two value
       if (rounding !== null) {
@@ -28,7 +30,7 @@ const computeLabel = (translate, label, boundaries, index, itemCount, rounding =
     return parseFloat(lower.toFixed(rounding)).toLocaleString();
   }
   // No value
-  return translate('terralego.legend.label');
+  return translate('terralego.legend.nolabel');
 };
 
 export const Legend = ({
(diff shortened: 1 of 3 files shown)


src/modules/Map/InteractiveMap/components/Legend/components/Line.js
```diff
@@ -0,0 +1,15 @@
...
+
+export const Line = ({ color, size, strokeWidth = 2 }) => (
+  <div className="tf-legend__symbol">
+    <svg
+      viewBox="0 0 16 16"
+      width={size}
+      height={size}
+    >
+      <line x1="0" y1="0" x2="16" y2="16" style={{ stroke: color, 'stroke-width': strokeWidth }} />
+    </svg>
```
dtrucs, line 11: If you want to keep the strokeWith in the same value whatever the size, you also need to change the value of the viewBox with the size  

```suggestion
    <svg
      viewBox={`0 0 ${size} ${size}`}
      width={size}
      height={size}
    >
      <line x1="0" y1="0" x2={size} y2={size} style={{ stroke: color, 'stroke-width': strokeWidth }} />
    </svg>
```

dtrucs: And set a default value in size to 16 line.3

src/modules/Map/InteractiveMap/components/Legend/Legend.js
```diff
@@ -106,23 +110,21 @@ export const Legend = ({
...
               <div
                 className="tf-legend__symbol-container"
                 style={{ width: shape === 'circle' ? biggestDiameter : DEFAULT_DIAMETER * 2 }}
               >
-                {shape === 'circle'
-                  ? <Circle color={color} size={diameter} />
-                  : <Rect color={color} size={DEFAULT_DIAMETER} />}
+                {(shape === 'circle') && <Circle color={color} size={diameter} />}
+                {(shape === 'line') && <Line color={color} size={DEFAULT_DIAMETER} strokeWidth={strokeWidth} />}
+                {(shape === 'square') && <Rect color={color} size={DEFAULT_DIAMETER} />}
```
mabhub, line 127: Another way to write this could be:
*(I don't know if it's worth it or not. Probably not)*
```suggestion
                {{
                  circle: <Circle color={color} size={diameter} />,
                  line:  <Line color={color} size={DEFAULT_DIAMETER} strokeWidth={strokeWidth} />,
                  square:  <Rect color={color} size={DEFAULT_DIAMETER} />,
                }[shape]}
```

jrmi: The problem is the error if the configuration contains error and the shape is bad. With previous version there is no display with the new one we have an error.

src/modules/Map/InteractiveMap/components/Legend/components/Line.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+export const Line = ({ color, size, strokeWidth = 2 }) => (
```
dtrucs, line 3: You really should set a default value to size 🙃

jrmi: Done

PR #190, opened 2020-04-03
Update linting

mabhub:


src/modules/Map/InteractiveMap/InteractiveMap.test.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable max-classes-per-file */
```
dtrucs, line 1: oh god, lazy boys (including me) would not like this rule :D

src/modules/Forms/Controls/Select/Select.js
```diff
@@ -125,8 +125,7 @@ export class Select extends React.Component {
             onQueryChange={this.handleQueryChange}
             initialContent={(values && values.length > 250)
               ? <MenuItem disabled text={translate('terralego.forms.controls.select.select_placeholder')} />
-              : undefined
-            }
+              : undefined}
```
dtrucs, line 128: Not the subject of the PR, but  `&&` seems more appropriate than a conditional ternary

src/modules/Map/InteractiveMap/InteractiveMap.test.js
mabhub: This is a test file.
If you want to group tests by concern, you have to allow exceptions to this rule. (Tests requires creating temporary classes)

src/modules/Forms/Controls/Select/Select.js
mabhub: Right, but as you said, I try to not embed in this PR anything else than linting ruleset upgrade.

PR #193, opened 2020-04-07
Fix selector again

jrmi:


src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeExclusiveItemsList/Selector/Selector.js
```diff
@@ -8,81 +8,132 @@ import translateMock from '../../../../../../utils/translate';
...
+}) => {
+  const [noMatchingLayer, setNoMatchingLayer] = React.useState(false);
+
+  const getLayerForSelector = React.useCallback(currentSelectorKey =>
+    layers.findIndex(({ selectorKey: layerSelectorKey }) =>
       Object.keys(layerSelectorKey).length === selectors.length &&
       selectors.reduce((prev, { name: selectorName }) =>
-        prev && layerSelectorKey[selectorName] === newKey[selectorName],
-      true));
+        prev && layerSelectorKey[selectorName] === currentSelectorKey[selectorName],
```
dtrucs, line 65: To be tested, but it think you can simplify this reduce
```js
selectors.every({ name: selectorName }) => layerSelectorKey[selectorName] === currentSelectorKey[selectorName])
```

Probably even better with `some` method and check a diff and win less iterations

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeExclusiveItemsList/Selector/Selector.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
```
mabhub, line 1: Why using name import for `useEffect` when all other React hooks are used through `React` object?

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeExclusiveItemsList/Selector/Selector.js
```diff
@@ -8,81 +8,132 @@ import translateMock from '../../../../../../utils/translate';
...
+  /**
+   * Get selected value from all values.
+   */
+  const getSelectedValue = React.useMemo(() => {
+    if (values.length) {
+      const found = values.find(({ value }) => currentValue === value);
+      if (!found) {
+        return values[0].value;
+      }
+      return found.value;
```
dtrucs, line 33: Not sure if this is better, but it could be simplify like this
```js
const [{ value } = values[0]] = values.filter(({ value }) => currentValue === value);
return value
```

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeExclusiveItemsList/Selector/Selector.js
```diff
@@ -8,81 +8,132 @@ import translateMock from '../../../../../../utils/translate';
...
+  }, [getLayerForSelector, onChange, selectorKey]);
+
+  useEffect(() => {
+    const activeLayerIndex = getLayerForSelector(selectorKey);
+
+    if (activeLayerIndex === -1) {
+      setNoMatchingLayer(true);
+    } else {
+      setNoMatchingLayer(false);
+    }
```
dtrucs, line 89:
```suggestion
  setNoMatchingLayer(activeLayerIndex === -1);
```

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeExclusiveItemsList/Selector/Selector.js
mabhub: `.filter()` will iterate over the entire array, when `.find()` stop as soon as callback return a truthy value...

dtrucs: True 😉

mabhub: This would achieve the same and also keep the `.find()`

```suggestion
      const { value } = values.find(({ value }) => currentValue === value) || values[0];
      return value;
```

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeExclusiveItemsList/Selector/Selector.js
jrmi: Indeed. Fixed.

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeExclusiveItemsList/Selector/Selector.js
jrmi: Done

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeExclusiveItemsList/Selector/Selector.js
jrmi: Done

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeExclusiveItemsList/Selector/Selector.js
jrmi: Done

src/modules/Forms/Controls/MultiSelect.js
```diff
@@ -75,10 +75,13 @@ export class MultiSelect extends React.Component {
...
-    const formatedValue = value.map(val => ({
-      label: formatedValues.find(({ value: availableVal }) => val === availableVal).label,
-      value: val,
-    }));
+    const formatedValue = value.map(val => {
+      const found = formatedValues.find(({ value: availableVal }) => val === availableVal);
+      return {
+        label: found ? found.label : '',
+        value: val,
+      };
```
mabhub, line 83:
```suggestion
      const { label } = formatedValues.find(({ value: availableVal }) => (val === availableVal)) || { label: '' };
      return {
        label,
        value: val,
      };
```

jrmi: done

PR #200, opened 2020-05-18
More consistent way of handling report url

CFarcy:


src/modules/Map/Map/components/ReportControl/ReportControl.js
```diff
@@ -20,36 +23,56 @@ export default class ReportControl extends AbstractControl {
...
+
+  setReportMarker = () => {
+    const currentHash = window.location.href.split('#')[1];
+    const fromReport = parse(currentHash).report;
+    if (!fromReport) {
+      return;
+    }
+
+    const { map } = this.props;
+    const mapHash = parse(currentHash).map;
```
dtrucs, line 50: 1/ You can directly access to the hash of the location with `location.hash` so you can refacto currentHash by
`const currentHash = location.hash.slice(1)`

2/ You might not need to do this step because `parse` from query string package would remove the `#` character so
`const fromReport = parse(location.hash).report;`

3/ So you can refacto all the things by destructuring:
```jsx
    const { report, map: mapHash } = parse(location.hash);
    if (!report) {
      return;
    }
```

src/modules/Map/Map/components/ReportControl/ReportControl.js
```diff
@@ -20,36 +23,56 @@ export default class ReportControl extends AbstractControl {
...
+    const [lat, lng] = mapHash.split('/').splice(1, 2); // do not need the zoom at index 0
+    if (!this.marker) {
+      this.marker = new Marker();
+    }
+    this.marker
+      .setLngLat({ lat: parseFloat(lat), lng: parseFloat(lng) })
```
dtrucs, line 56:
```js 
.setLngLat({ lat: Number(lat), lng: Number(lng) })
```

Not mandatory but for readability purpose

src/modules/Map/Map/components/ReportControl/ReportControl.js
```diff
@@ -20,36 +23,52 @@ export default class ReportControl extends AbstractControl {
...
+  }
+
+  setReportMarker = () => {
+    const { report, map: mapHash } = parse(window.location.hash);
+    if (!report) {
+      return;
+    }
+    const { map } = this.props;
+
+    const [lat, lng] = mapHash.split('/').splice(1, 2); // do not need the zoom at index 0
```
jrmi, line 47: You may do that:
```suggestion
    const [_zoom, lat, lng] = mapHash.split('/');
```

dtrucs:
```js  
const [🔍, lat, lng] = mapHash.split('/');
```
🤓

CFarcy: I used `_zoom` but if it's better to use the emoji I can change it. Initially, I used `splice` to please the linter. So I just disable `no-unused-vars` for this line.

PR #202, opened 2020-06-04
Improve main menu

dtrucs:


src/components/MainMenuItem.js
```diff
@@ -19,7 +19,7 @@ const MainMenuItem = ({ component, isMobileSized, id, ...props }) => {
 };
 
 MainMenuItem.propTypes = {
-  component: PropTypes.node,
+  component: PropTypes.func,
```
jrmi, line 22: Is it always a func ?

dtrucs: in fact, this is always a component, but console warns with de `propTypes.node`. Maybe there is a bette, explicite, propTypes for component?

dtrucs: @jrmi maybe i could set 

```suggestion
  component: PropTypes.oneOfType([PropTypes.element, PropTypes.func]),
```

If i remove func, in some cases there is a warn.

PR #203, opened 2020-06-05
Improve login button component

dtrucs:


src/components/LoginButton/LoginButton.js
```diff
@@ -66,8 +68,8 @@ export const LoginButton = props => {
               title={t('auth.loginform.title')}
               panel={(
                 <LoginForm
-                  onBeforeSubmitting={onBeforeSubmitting}
-                  onAfterSubmitting={onAfterSubmitting}
+                  onBeforeSubmitting={() => setIsNewlyAuthenticated(true)}
```
jrmi, line 71: Why do you remove useCallback here ?

dtrucs: When there are no dependencies to props, useCallback are useless and even must be worse because it _is doing more work_. (more explanation here https://kentcdodds.com/blog/usememo-and-usecallback)

PR #210, opened 2020-07-03
Refacto story component

dtrucs:


src/modules/Visualizer/Story/Story.js
```diff
@@ -20,123 +21,108 @@ const resetLayers = map => ({ layers = [] }) => {
...
-  state = {
-    step: 0,
-  }
+export const Story = ({ map, story: { beforeEach = [], slides }, setLegends, translate }) => {
+  const [step, setStep] = useState(0);
 
-  componentDidMount () {
-    this.initMap();
-  }
+  const slidesLength = slides.length - 1;
```
mabhub, line 27:
```suggestion
  const slidesLength = React.useMemo(() => (slides.length - 1), [slides.length]);
```

src/modules/Visualizer/Story/Story.js
```diff
@@ -20,123 +21,108 @@ const resetLayers = map => ({ layers = [] }) => {
...
+    setStep(prevStep => (prevStep === slidesLength
       ? 0
-      : prevStep + 1 }));
-  }
+      : prevStep + 1));
+  }, [slidesLength]);
 
-  initMap () {
-    const { map } = this.props;
+  const { title, content } = slides[step];
```
mabhub, line 47:
```suggestion
  const { title, content } = React.useMemo(() => slides[step], [slides, step]);
```

mabhub: I'm not really sure if this one is legit or not…

src/modules/Visualizer/Story/Story.js
dtrucs: With your suggestion, we allocate another function with `slides.length` as argument to return... `slides.length - 1`.
It's not a huge calculation and i am not sure if it's better for performance purposes.
What do you think @jrmi ?

mabhub: This insure that `slidesLength` won't change between renders when slides do not change…

But ultimately, as it is just a Number, it does not really matter.

So yep, you can forget about that ;)

jrmi: No need useMemo here, a number don't change is reference ! The useMemo is only usefull for high computation which is no the case here.

src/modules/Visualizer/Story/Story.js
jrmi: No high computation here also so don't need.

src/modules/Visualizer/Story/Story.js
jrmi: My problem is more the naming here. Are we really talking about length of a silde ? I propose `slideCount` instead.
```suggestion
  const slideCount = slides.length - 1;
```

PR #211, opened 2020-07-10
Improve LayersTreeItemFilters component

dtrucs:


src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeItemFilters/LayersTreeItemFilters.js
```diff
@@ -1,64 +1,136 @@
...
+  return value;
 }
 
 const LayersTreeItemFilters = ({
   filtersValues,
   setLayerState,
   layer,
-  layer: { filters: { form: layerForm } = {} } = {},
-  activeLayer: { filters: { form = layerForm } = {} } = {},
+  layer: { filters: { form: layerForm } = {} },
```
jrmi, line 46: Why did you remove the last `= {}` ?

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeItemFilters/LayersTreeItemFilters.js
```diff
@@ -1,64 +1,136 @@
...
+            onRemove={() =>
+              setLayerState({
+                layer,
+                state: {
+                  filters: { ...filtersValues, [property]: undefined },
+                  total: undefined,
+                },
+              })}
+          >
+            {label && translate('terralego.visualizer.layerstree.itemFilters.label', { label })}
```
jrmi, line 78: I'm not sure we can/should translate the label here ?

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeItemFilters/LayersTreeItemFilters.js
```diff
@@ -1,64 +1,136 @@
...
+  if (value instanceof Date) {
+    return translate('terralego.visualizer.layerstree.itemFilters.date', { value: getLiteralValue(value) });
   }
-  return `: ${getLiteralValue(value)}`;
+
+  if (type === 'many') {
+    return value.join(', ');
+  }
+
+  return value;
```
mabhub, line 39: - Prefer simple returns *early*.
- Negating cases may be simplier than nesting conditions

```jsx
export function getValueFromType (value, type, translate) {
  const isDate = val => val instanceof Date;
  const nameSpace = 'terralego.visualizer.layerstree.itemFilters';

  if (type === 'boolean') {
    return null;
  }

  if (type === 'many') {
    return value.join(', ');
  }

  if (isDate(value)) {
    return translate(`${nameSpace}.date`, { value: getLiteralValue(value) });
  }

  if (type !== 'range') {
    return value;
  }

  const [from, to] = value;

  if (!isDate(from) && !isDate(to)) {
    return translate(`${nameSpace}.range.numeric`, { from, to });
  }

  if (from && to) {
    return translate(`${nameSpace}.range.date`, { from: getLiteralValue(from), to: getLiteralValue(to) });
  }

  return translate(`${nameSpace}.date`, { value: getLiteralValue(from) });
};
```

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeItemFilters/LayersTreeItemFilters.js
dtrucs: Because the defaultProps is now defined at the bottom of the file with the propTypes

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeItemFilters/LayersTreeItemFilters.js
dtrucs: We need it for the space before the colon (or not) here https://github.com/Terralego/terra-front/pull/211/files#diff-29fe5a41570d779192152d9369cd7676R181 and here https://github.com/Terralego/terra-front/pull/211/files#diff-24c5ed2b0c1960ed409a573467e62accR130

src/modules/Visualizer/LayersTree/LayersTreeItem/LayersTreeItemFilters/LayersTreeItemFilters.js
dtrucs: thx 👍 .
I took it and i improve it for the case when the user is picking the begin or the end of the range date but not the other one.

PR #216, opened 2020-07-20
Fix legend auto

jrmi:
- Add label for null value
- Change last and first label to open ended version


jrmi:
Just take a look at the two last commit if you want to see difference without style modifications.

PR #217, opened 2020-07-23
SourceLayer is not mandatory for highlight

CFarcy:


src/modules/Map/InteractiveMap/InteractiveMap.js
```diff
@@ -514,13 +514,16 @@ export class InteractiveMap extends React.Component {
...
+          const highlightLayer = {
             id: highlightTypeId,
             source,
             type: highlightType,
-            'source-layer': sourceLayer,
             paint: targetType()[highlightType],
-          });
+          };
+          if (sourceLayer) {
+            highlightLayer['soure-layer'] = sourceLayer;
```
jrmi, line 524: Typo ? 
```suggestion
            highlightLayer['source-layer'] = sourceLayer;
```

PR #242, opened 2021-02-11
Fix stroke only square legend

jrmi:


src/modules/Map/InteractiveMap/components/Legend/components/Rect.js
```diff
@@ -2,15 +2,16 @@ import React from 'react';
...
   <div className="tf-legend__symbol">
-    <svg
-      viewBox={`0 0 ${size} ${size}`}
-      width={size}
-      height={size}
-    >
+    <svg viewBox={`0 0 ${size} ${size}`} width={size} height={size}>
       <rect
-        fill={color}
+        fill={color || '#00000000'}
```
dtrucs, line 14: Why don't you set the fallback in the `defaultProps` line.34?

dtrucs: And `transparent` could be set instead of `'#00000000' ?
It's more explicit imo

jrmi: sometimes value is set to null. But i may search why the value is null...

dtrucs: The defaultprops set to null 👀

jrmi: Fixed

jrmi: > The defaultprops set to null 👀

Yes i missed it 🤣 thanks. I always forget defailtProps...

jrmi: fixed with transparent.

dtrucs (approved):

PR #253, opened 2021-03-25
Merge layer filter instead of overriding it and apply it to geojson also

CFarcy:


src/modules/Map/services/cluster.js
```diff
@@ -85,7 +86,7 @@ export const createClusterLayers = ({
     ...layerProps,
     id: `${id}-${PREFIX_UNCLUSTERED}-${index}`,
     source: clusterSourceName,
-    filter: ['!', ['has', 'point_count']],
+    filter: ['all', ['!', ['has', 'point_count']], filter],
```
jrmi, line 89: What if filter is undefined ?

jrmi (changes requested):

src/modules/Map/services/cluster.js
CFarcy: `filter` is now set as an empty list by default

CFarcy: if `filter` is falsy, it is not merge with the rest

jrmi (approved):

PR #272, opened 2021-06-08
Do not render template with no data

CFarcy:


src/modules/Template/Template/Template.js
```diff
@@ -21,7 +21,7 @@ export const Template = ({
   ...props
 }) => {
   // Missing data could raise template render error
-  if (loading) {
+  if (loading || loading === undefined) {
```
jrmi, line 24: Can we find a way to avoid this strange test by modifying the parent component ?

CFarcy: Done, but I'm not completely satisfy with the solution. Even tough the error is gone, template is empty in some cases. 
I'm working on it!

CFarcy: I test it with the steps Marine gave me to reproduce the bug and it seems it's fixed.

PR #273, opened 2021-06-10
Remove previous interaction listeners if `interactions` prop did update

dtrucs:


CFarcy (approved):

src/modules/Map/services/mapUtils.js
```diff
@@ -181,20 +185,23 @@ export function setInteractions ({ map, interactions, callback }) {
 
     const { interactions: filteredInteractionsSpec, feature, layerId } = interactionsSpec;
     filteredInteractionsSpec.forEach(interaction =>
-      callback({ event: e, map, layerId, feature, interaction, eventType }));
+      callback({ event, map, layerId, feature, interaction, eventType }));
   };
+
+  const debounceListener = debounce(
```
jrmi, line 191: Why do you add a debounce here ?

src/modules/Map/services/mapUtils.test.js
```diff
@@ -10,6 +10,8 @@ import {
   PREV_STATE,
 } from './mapUtils';
 
+jest.mock('lodash.debounce', () => jest.fn(fn => (...args) => fn(...args)));
```
jrmi, line 13: Did you add a test to verify event handler are actually removed then re-added ?

src/modules/Map/services/mapUtils.js
dtrucs: I add it to replace the hacky debouncer with setInterval/clearInterval

src/modules/Map/services/mapUtils.test.js
dtrucs: I just add some tests related to it 👍

jrmi (approved):

PR #274, opened 2021-06-21
Catch template rendering error early to avoid app crashing

CFarcy:


jrmi (approved):

jrmi (dismissed):

jrmi (approved):

src/components/ErrorBoundary.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+export default class ErrorBoundary extends React.Component {
+  constructor (props) {
+    super(props);
+    this.state = { hasError: false };
+  }
```
dtrucs, line 7: Why not simply?
```suggestion
  state = { hasError: false };
```

src/modules/Template/Template/Template.js
```diff
@@ -23,36 +23,47 @@ export const Template = ({
     ? env.renderString(template, props)
     : content);
 
-  const processNodeDefinitions = new ProcessNodeDefinitions(React);
+  const processingInstructions = useMemo(() => {
+    const processNodeDefinitions = new ProcessNodeDefinitions(React);
+    const customInstructions = customComponents.map(spec => {
+      if (spec.processNode) {
+        return spec;
+      }
```
dtrucs, line 31: Do you really want a `customInstructions` array with potential holes? Maybe a `filter` method should be more appropriate?

src/components/ErrorBoundary.js
CFarcy: I just copy/past the doc https://reactjs.org/docs/error-boundaries.html, but I can change it if not needed

src/modules/Template/Template/Template.js
CFarcy: It either return the `spec` or an object construct around `spec`

jrmi: I don't see where is the hole neither?

dtrucs: wow that's my bad, at first glance I thought the `map` ended after that `if`.

dtrucs (approved):

PR #327, opened 2023-06-05
Add sso login button if url is available

CFarcy:


src/modules/Auth/components/AuthProvider/index.js
```diff
@@ -45,9 +46,12 @@ export class AuthProvider extends React.Component {
     await createToken(properties);
   }
 
-  logoutAction = () => {
+  logoutAction = async (ssoLink = null) => {
     clearToken();
     this.setState({ authenticated: false, user: null });
+    if (ssoLink) {
+      await Api.request(ssoLink);
+    }
```
mabhub, line 54: Maybe should we call logout link for sso _before_ clearing token?

mabhub (approved):

src/modules/Auth/components/AuthProvider/index.js
CFarcy: done

mabhub (approved):

PR #341, opened 2023-06-16
Add uuid to Legends to use as key

Bo-Duke:
Fixes bug where legends without title could be duplicated.

Also removes logic which enables zoom restrictiction on legends as it isn't currently used

### DIFF
--- a/src/modules/Map/InteractiveMap/InteractiveMap.js
+++ b/src/modules/Map/InteractiveMap/InteractiveMap.js
@@ -168,7 +168,6 @@ export class InteractiveMap extends React.Component {
         ? backgroundStyle[0].url
         : backgroundStyle,
       legends: [],
-      uuidLegends: [],
       interactionList: [],
     };
   }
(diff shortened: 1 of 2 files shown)


src/modules/Map/InteractiveMap/InteractiveMap.js
```diff
@@ -296,17 +293,13 @@ export class InteractiveMap extends React.Component {
...
-    const { map } = this;
-    if (!map) return;
-    const zoom = map.getZoom();
-    const filteredLegends = legends
-      .filter(({ minZoom = 0, maxZoom = Infinity }) =>
-        ((zoom === 0 && minZoom === 0) || zoom > minZoom) && zoom <= maxZoom);
-
-    this.setState({ legends: filteredLegends });
-  };
+    const uuidLegends = legends.map(legend => ({ ...legend, renderUuid: uuid() }));
```
Bo-Duke, line 298: Yeah all new legends are rerendered when one is added / removed because there's no easy way to ID unique legends. It's good enough and still better than before.

CFarcy (approved):

mabhub (approved):

PR #345, opened 2023-06-27
Fix pie-charts opacity and radius

Bo-Duke:


src/modules/Map/Map/Map.js
```diff
@@ -435,13 +454,19 @@ export class MapComponent extends React.Component {
...
     };
 
     map.on('render', () => {
       if (!map.isSourceLoaded(layer.source)) return;
-      const { fields, chart_radius: chartRadius } = layer.advanced_style;
+      const { fields, ...advancedStyle } = layer.advanced_style;
       const usableFields = fields.filter(field => field.use);
-      updateMarkers(usableFields, chartRadius);
+      updateMarkers(usableFields, advancedStyle);
     });
```
mabhub, line 470: Do we know how often this event occurs?

mabhub (approved):

src/modules/Map/Map/Map.js
Bo-Duke: Quite often.

Nothing is made if the features on screen haven't changed but yeah, there's a map.queryRenderedFeatures each time the user pan or zooms.

I have lots of ideas to improve performances, especially in cases where we have several piechart layers in the same view but I'll do it with the upcoming radius change (radius of piechart being calculated in the back instead of at runtime)

PR #349, opened 2023-07-07
Add filter input to layerstree

Bo-Duke:
Still some things to finish up, especially about the style. (And if the feature should be hidden behind a button)

### DIFF
--- a/src/modules/Map/Map/components/SearchControl/SearchInput.js
+++ b/src/modules/Map/Map/components/SearchControl/SearchInput.js
@@ -2,19 +2,14 @@ import React from 'react';
 import { Card, InputGroup, Button } from '@blueprintjs/core';
 
 export const SearchInput = ({
-  onChange,
   query,
   onClose,
   onKeyPress,
-  onFocus,
   loading,
-  placeholder,
+  ...props
 }) => (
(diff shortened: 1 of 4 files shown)


src/modules/Map/Map/components/SearchControl/SearchInput.js
```diff
@@ -8,9 +8,11 @@ export const SearchInput = ({
   onKeyPress,
   onFocus,
   loading,
+  placeholder,
 }) => (
   <Card className="search-input">
     <InputGroup
+      placeholder={placeholder}
```
mabhub, line 15: What about:

```suggestion
  ...rest
}) => (
  <Card className="search-input">
    <InputGroup
      {...rest}
```

or maybe something like:

```js
export const SearchInput = ({
  query: value,
  onClose,
  loading,
  cardProps = {},
  ...props
}) => (
  <Card className="search-input" {...cardProps}>
    <InputGroup
      rightElement={(
        <Button
          icon="small-cross"
          minimal
          loading={loading}
          onClick={onClose}
        />
      )}
      {...props}
    />
  </Card>
);
```

mabhub (approved):

